PR #10, opened 2020-06-11
Feature/error handler

lagartoverde:
Added an error handler, so in the future we can send the errors to the backend for storing them


app/instance-initializers/error-handler.js
```diff
@@ -0,0 +1,31 @@
+export function initialize(appInstance) {
+  const store = appInstance.lookup('service:store');
+  
+  window.onerror = async function (msg, url, lineNo, columnNo, error) {
+    const logSource = await store.query('log-source', {
```
nvdk, line 5: why not use ```store.findRecord``` ?

app/models/custom-inflector-rules.js
```diff
@@ -59,6 +59,7 @@ inflector.irregular('task', 'tasks');
 inflector.irregular('tasklist-solution', 'tasklist-solutions');
 inflector.irregular('task-solution', 'task-solutions');
 inflector.irregular('blockchain-status', 'blockchain-statuses');
+inflector.irregular('log-entry', 'log-entries');
```
nvdk, line 62: you'll need to add log-source and log-level here as well

nvdk (approved):

nvdk:
closing due to inactivity, looks like we won't use this

PR #13, opened 2020-07-02
Feature/fix bug with untittled documents

lagartoverde:


app/components/editor-document-title.js
```diff
@@ -41,7 +44,7 @@ export default Component.extend({
         schedule('afterRender', () => this.$('input').focus());
       else {
         this.set('overruledTitle', this.editorDocument.title);
-        if (this.generatedTitle.length > 0 && this.overruledTitle.length === 0)
+        if (this.generatedTitle.length > 0 && this.overruledTitle.trim().length === 0)
```
nvdk, line 47: trim actually modifies the existing string, we probably don't want that

PR #15, opened 2020-10-12
Feature/besluit editor and behandeling fields

lagartoverde:
All of https://binnenland.atlassian.net/browse/GN-1745


nvdk (changes requested):
In general seems to work, I was a bit confused because you had code attempting to set properties such as folder, status and title while those were never provided. It did lead to an interesting discussion, so please make sure to set the title, folder and status as we discussed:
- title can match the agendapoint title
- folder can be the drafts folder for now
- the status should be active or archived.

It seems there was some misunderstanding between me and felix on the use of i18n, you can and should use it where possible. 

Naming things is important, specifically for the profile of the editor a better name than "newFlowProfile" should be possible. We should also consider whether we can not just reuse the already existing drafts profile.

Finally I think we should really be more defensive in the saving of the document, currently there is no error handling at all and we're actually fetching and saving a lot of data. at the very least we should have a try/catch and inform the user if something went wrong. I think I'll create a separate ticket for that,

app/config/editor-profiles.js
```diff
@@ -77,5 +77,15 @@ export default {
     "rdfa-editor-roadsign-hint-plugin"
 
   ],
+  newFlowProfile: [
```
nvdk, line 80: please provide a decent name for the profile, also consider whether it can be merged with the drafts profie

app/components/behandeling-van-agendapunt.js
```diff
@@ -1,7 +1,96 @@
...
+      yield this.document.save();
+      yield this.documentContainer.save();
+      yield this.behandeling.save();
+    }
+  }
+  @action
+  async save(e) {
+    e.stopPropagation();
+    this.behandeling.openbaar = this.openbaar;
+    await this.saveEditorDocument.perform(this.document);
```
nvdk, line 45: we should at least have some error handling here and not update the behandeling if the save of the document failed

app/components/behandeling-van-agendapunt.js
```diff
@@ -1,7 +1,96 @@
...
+    this.editor = editor;
+  }
+
+  @action
+  toggleOpenbaar(e) {
+    this.openbaar = e.target.checked;
+  }
+
+  @task
+  *saveEditorDocument(editorDocument, newStatus, newFolder) {
```
nvdk, line 63: this could be more defensive

nvdk (approved):

PR #16, opened 2020-10-12
Add participation and voting management to the behandeling van agendapunt

abeforgit:
First of all, this should have been 2 (or even 3) pull requests, apologies for that. 

## Relevant jira tickets:
https://binnenland.atlassian.net/browse/GN-1776
https://binnenland.atlassian.net/browse/GN-1778
https://binnenland.atlassian.net/browse/GN-1779

## Remarks
- changing the participants of a behandeling does not affect participants of existing elections
- currently any attendee who does not vote is an "onthouding", there is no distinction whether the attendee was not allowed to vote or if they chose not to, but this can be added easily
- there is still a visual power-select bug, which may or may not be my fault
- slightly depends on #15 in that the components exist inside the container, but nothing more than that


app/components/agenda/edit.js
```diff
@@ -1,11 +1,11 @@
...
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+import { action } from "@ember/object";
 
 export default class AgendaEditComponent extends Component {
   @action
-  async toggleGeplandOpenbaar(){
-    this.args.agendapunt.geplandOpenbaar=!this.args.agendapunt.geplandOpenbaar;
+  async toggleGeplandOpenbaar() {
+    this.args.agendapunt.geplandOpenbaar = !this.args.agendapunt
```
nvdk, line 8: nitpicking, but I really liked this better on one line, perhaps increase minimum line width in your editor a bit?

app/components/meeting-form.js
```diff
@@ -1,24 +1,42 @@
...
     this.voorzitter = this.args.zitting.get('voorzitter');
     this.aanwezigenBijStart = this.args.zitting.get('aanwezigenBijStart');
+    this.fetchBehandelingen.perform();
+
+  }
+
+  /** @type {import("ember-concurrency")} */
+  @task
+  fetchBehandelingen = function*() {
+    /** @type {import("../models/agendapunt").default[]} */
```
nvdk, line 29: importing the typedefs first like you do in app/components/treatment/participation.js makes this more readable.

app/components/agenda/modal.js
```diff
@@ -11,30 +11,55 @@ export default class AgendaModalComponent extends Component {
 
   @tracked currentlyEditing;
 
+  isNew = false;
+  get afterSave() {
```
nvdk, line 15: I prefer an empty line between class fields and methods, makes it a bit more readable (but again nitpicking)

app/components/treatment/voting/voter-row.hbs
```diff
@@ -0,0 +1,30 @@
...
+  <td>
+    {{this.persoon.gebruikteVoornaam}} {{this.persoon.achternaam}}
+  </td>
+  <td>
+    {{@row.[0].bekleedt.bestuursfunctie.label}}
+  </td>
+  <td>
+    {{#if @manageVotes}}
+    <PowerSelect @searchEnabled={{false}} @options={{this.voteOptions}} @selected={{this.selectedVote}}
+      @onchange={{this.onSelectVote}} as |singleselect|>
```
nvdk, line 12: You can fix the PowerSelect bug by using "onChange" instead of "onchange" I believe.
```
<PowerSelect @searchEnabled={{false}} @options={{this.voteOptions}} @selected={{this.selectedVote}}
      @onChange={{this.onSelectVote}} as |singleselect|>
```

nvdk (changes requested):
Some minor bugs, but all in all seems to do what it should do, however we really need to resolve the select bug as the voting component isn't really usable in the current state. given the limited use of powerselect functionality I would recommend using a regular select which will also be more accessible.

general remarks:
- after managing votes "onderwerp" shows " [object Object] " in the overview table when left empty
- after managing votes "gevolg" shows " [object Object] " in the overview table when left empty
- pressing save when managing votes seems to also seems to trigger a save of each behandeling van agendapunt, I don't see why at the moment
- the select dropdown not hiding after select eally needs to be fixed. it seems as if the select box is actually rendered twice after select.
- order of treatments is not inline with actual agenda (but I think not in scope for this PR?)

app/components/treatment/voting/voter-row.hbs
abeforgit: I tried onChange first but it didnt work
Ill try to see if the WU version works better

nvdk:
encountered another bug, when you edit an existing vote it seems to longer load the mandatarissen

nvdk:
@abeforgit since I've merged oscar's PR can you also resolve the conflicts?

abeforgit:
Sorry for the late-night response, I had a lot of unplanned errands to run during the day
>     * pressing save when managing votes seems to also seems to trigger a save of each behandeling van agendapunt, I don't see why at the moment

I can't reproduce this, maybe this was related to not having the changes to the backend yet?

>     * order of treatments is not inline with actual agenda (but I think not in scope for this PR?)

No, for that I think its best to wait for https://binnenland.atlassian.net/browse/GN-1781

> encountered another bug, when you edit an existing vote it seems to longer load the mandatarissen

This is because on edit it needs to fetch the voter list from the backend. On create, we are making a new stemming object so we already know all the voter information. As a solution we can prefetch all the voter information in the query for the  behandelingen with includes, but that will make that query quite heavy when there are a lot of agendapunten

PR #18, opened 2020-10-15
added read only functionaly to the archived folder

Asergey91:
GN-1743


nvdk (approved):
needs i18n, but I think oscar was picking that up

PR #19, opened 2020-10-19
Feature/new published resource flow

abeforgit:
Fixes https://binnenland.atlassian.net/browse/GN-1811

Reuses a most of the old UI components, but updated to the octane standard.

related prs:
- [backend](https://github.com/lblod/app-gelinkt-notuleren/pull/7)
- [service](https://github.com/lblod/notulen-prepublish-service/pull/3)


app/components/signatures/agenda/timeline-step.js
```diff
@@ -1,50 +1,50 @@
...
-      });
-    }
-  })
-});
+  get mockAgenda() {
+    return {
+      body: this.args.agenda.renderedContent,
+      signedId: this.args.agenda.zitting.get('id')
+    };
+    //   if (this.args.agenda.renderedContent) {
```
nvdk, line 30: if this is no longer used, feel free to remove the commented out code. any specific reason to keep it here?

app/templates/meetings/publish.hbs
```diff
@@ -0,0 +1,45 @@
+<div class="container-flex--contain">
+  <div class="au-o-box au-o-box--small au-c-app-chrome">
+    <div class="u-spacer--tiny">
+      <LinkTo
+        @route="meetings.edit"
+        @model={{@model.id}}
+        class="smaller link-dark"
+      >
+        ← Terug naar notulen
```
nvdk, line 9: perhaps update to terug naar zitting ?

nvdk (approved):
Well done :100:

lagartoverde (approved):
Apart from Niels comments everything else looks good to me 👍

nvdk:
found another bug, if no zitting was created yet click sign and publish results in an error.

nvdk:
also throws an error if no agendapoints are linked to the zitting. this will need to be a bit more user friendly I'm afraid :). We can either check in the frontend or have the backend provide error messages that can be acted on. (the proper error code would probably be 400).

nvdk:
ah and should have been more clear in my earlier remark, but "besluit:behandelt" also needs to be nested under the zitting, because it's also linked to the zitting.

abeforgit:
Well to me it makes the most sense to just disable the publish button when these things are not present, no?

PR #22, opened 2020-10-23
Feature/new flow

nvdk:
This PR hopes to provide the first steps toward publishing an agenda in the new flow

TODO:
- [x] merge @Asergey91's sorting branch
- [x] make sure saving is consistent in the entire form


abeforgit:
So I went through the meeting-form with a fine comb with the primary aim of checking the saving behaviour (which, as far as I can tell, is actually consistent everywhere)
Ive also recorded any bugs/unexpected behavior I found. Heres a list, with the items in bold being the ones I think are the ones that are most relevant to this sprint and as such are the ones I will start working on immediately, let me know if any others take priority.


## Everywhere
- **modals should reset when cancelled and reopened**


## Aanwezigen
- why does opening the modal take so long when the attendees are already loaded?
- **mandataries -> mandatees**
- chairman and secretary selector should contain the current attendees without searching
- it should probably be impossible to have an absent chairman/secretary
- canceling and reopening should set the chairman and secretary selectors back to the "real selected" values
- **secretary selector doesn't work**
## Behandelingen
- **getting the attendees of the previous treatment should be possible now**
- **chairman and secretaris are not picked up in treatment attendees**
- do we really need the chairman and secretaris selectors for every treatment?

### Voting
- we probably want a warning when voting results dont match up in secret votings
- cancellation seems to work fine here, except for top-level voting properties (i.e. subject, consequence, etc)

nvdk:
short feedback:
> do we really need the chairman and secretaris selectors for every treatment? 
yes, because they can be temporarily replaced, for example if the treatment is connected to them personally they will need to leave the room. it is an exception though.

PR #26, opened 2020-10-26
Feature/agendapunt ordering

Asergey91:
stuff seems to be mostly tested and working

PR #25, opened 2020-10-23
Bugfix/modal cancelling

abeforgit:
- [x]  general information
- [x] Attendance
- [x] Agenda
- [x] Behandelingen

PR #27, opened 2020-10-29
Feature/gn demo branch

abeforgit:
This pr combines #25 and #26 and also has some additional i18n improvements


app/components/agenda/list.hbs
```diff
@@ -18,40 +18,42 @@
...
-                <i class="vi vi-edit"></i> {{t "manageAgendaZittingModal.editItemButton"}}
-              </WuButton>
-            {{/wu-button-group}}
-          </td>
-        {{/draggable-object}}
+        {{#unless agendapunt.isDeleted}}
+          {{#draggable-object tagName="tr" content=agendapunt isSortable=true}}
+            {{!--  <td class="u-align-right">1</td> --}}
+            <td>
+              {{#wu-link-button
```
nvdk, line 27: we should avoid usage of webuniversum (so `wu-` components) and prefer appuniversum where possible. This is also using pre octane syntax.
The equivalent in appuniversum would be
```
           <AuButton @skin="secondary"><span style="cursor:pointer"><i class="fa fa-bars"></i></span></AuButton>
```
I think

app/components/agenda/list.hbs
```diff
@@ -18,40 +18,42 @@
...
+            </td>
+            <td>
+              <p>{{agendapunt.titel}}</p>
+            </td>
+            <td>
+              {{if agendapunt.geplandOpenbaar (t "manageAgendaZittingModal.geplandOpenbaarTrueLabel") (t "manageAgendaZittingModal.geplandOpenbaarFalseLabel")}}
+            </td>
+            <td>
+              {{#wu-button-group}}
+                <WuButton
```
nvdk, line 45: same as above, although this is using octane syntax, somewhat sloppy mixing them ;)

app/components/agenda/modal.hbs
```diff
@@ -1,14 +1,32 @@
 {{#wu-modal
```
nvdk, line 1: If possible we should use an appuniversum equivalent of wu-modal. if it doens't exist, we should consider creating one.

app/components/agenda/move.hbs
```diff
@@ -0,0 +1,38 @@
+<label class="form__label">Verplaats agendapunt
+  <div class="grid">
+    <div class="col--6-12">
+      {{#power-select
```
nvdk, line 4: this should be converted to octane syntax, where possible (and I believe here it's very possible) it's better to avoid power select if we don't need advanced features (like search or typeahead). For a simple select, prefer the select element and element modifiers.

app/components/agenda/move.hbs
```diff
@@ -0,0 +1,38 @@
...
+           selected=@selectedLocation
+           allowClear=true
+           onchange=@selectLocation
+           as |location|}}
+        {{location.name}}
+      {{/power-select}}
+    </div>
+    {{#if @showAfterAgendapuntOptions}}
+      <div class="col--6-12">
+        {{#power-select
```
nvdk, line 20: here we could use power select, because we offer type ahead. but it should definitely be converted to octane syntax.

app/components/agenda/delete.js
```diff
@@ -12,11 +12,10 @@ export default class AgendaDeleteComponent extends Component {
     this.isShowingWarning=!this.isShowingWarning;
   }
 
-  @action
-  async delete(){
-    await this.args.agendapunt.destroyRecord();
-    this.args.toggleEditing();
-    this.args.afterSave();
-  }
+  // @action
```
nvdk, line 15: commented out code should just be removed, if there is a reason to keep it it should be included in the comment block

nvdk (changes requested):

app/components/agenda/move.hbs
```diff
@@ -0,0 +1,38 @@
+<label class="form__label">Verplaats agendapunt
+  <div class="grid">
+    <div class="col--6-12">
+      {{#power-select
+           class="select select--block u-spacer--tiny"
+           renderInPlace=true
+           placeholder="Selecteer positie[@TN]"
```
nvdk, line 7: this is not using a i18n string and should be added to our translations

app/components/agenda/move.hbs
```diff
@@ -0,0 +1,38 @@
...
+        {{location.name}}
+      {{/power-select}}
+    </div>
+    {{#if @showAfterAgendapuntOptions}}
+      <div class="col--6-12">
+        {{#power-select
+             class="js-select u-spacer--small"
+             renderInPlace=true
+             placeholder="Selecteer agendapunt"
+             searchPlaceholder="Zoek op titel"
```
nvdk, line 24: this is not using a i18n string and should be added to our translations

app/components/agenda/move.hbs
```diff
@@ -0,0 +1,38 @@
...
+           as |location|}}
+        {{location.name}}
+      {{/power-select}}
+    </div>
+    {{#if @showAfterAgendapuntOptions}}
+      <div class="col--6-12">
+        {{#power-select
+             class="js-select u-spacer--small"
+             renderInPlace=true
+             placeholder="Selecteer agendapunt"
```
nvdk, line 23: this is not using a i18n string and should be added to our translations

app/components/agenda/list.hbs
Asergey91: It kinda breaks the design in this case hopefully thats fine

app/components/agenda/modal.hbs
abeforgit: It does exist, but it doesn't allow parent components to open/close it yet. Instead it renders it's own button, so it's pretty awkward to use atm, there is a TODO outstanding for this
https://appuniversum.github.io/ember-appuniversum/docs/components/au-modal
Will update it with octane syntax

app/components/agenda/move.hbs
abeforgit: It's been converted to octane so I will leave it for now
I'm not sure I agree with avoiding power-select as much as possible, I find simple selects annoying to work with and having two ways to deal with selects in the same component can be confusing.
Is powerselect really that heavy?

PR #28, opened 2020-11-06
Feature/besluitenlijst publication

lagartoverde:


app/components/signatures/besluitenlijst/timeline-step.js
```diff
@@ -1,36 +1,10 @@
...
-            .request(`/prepublish/besluitenlijst/${this.currentEditorDocument.id}`)
-            .then( (response) => get(response, "data.attributes.content") ),
-          signedId: get( this, 'currentEditorDocument.id' )
-        })
-      });
-    }
-  })
-});
+export default class SignaturesBesluitenlijstTimelineStep extends Component {
+  get mockLijst() {
```
nvdk, line 4: I wonder why you no longer need the if/else branch in mockLijst

nvdk (commented):
Still need to test this, but looks quite ok!

app/components/signatures/besluitenlijst/timeline-step.js
lagartoverde: I just followed Arne code for the agenda, he does this on the controller level instead of the component, I thought it was good to be consistent with the code :)

PR #29, opened 2020-11-19
Feature/behandeling publication

lagartoverde:


app/controllers/meetings/publish/uittreksels.js
```diff
@@ -0,0 +1,73 @@
...
+          behandeling,
+        });
+        this.uittreksels.push(rslt);
+      }
+    }
+  }
+
+
+  @task
+  *createPrePublishedResource() {
```
nvdk, line 47: shouldn't this also get the behandeling as an argument?

lagartoverde: No, because this one gets the prepublished resource from the service, doesn't need the old behandeling if there is one

lagartoverde: I keep the naming consistent with the other routes, but maybe this one should be called getPrepublishedBehandelings
As it gets all of them not just one

nvdk: As discussed I think this might not work for larger agenda's (50+ agendapoints). perhaps better to work with a request per agendapoint and only do the request if that one was selected

nvdk:
So to summarize, I think we need to rework this a bit. I'd suggest first fetching what you need to display the titles and only fetching either the existing uittreksel or a preview when necessary.

app/controllers/meetings/publish/uittreksels.js
```diff
@@ -0,0 +1,74 @@
...
+  initialize() {
+    this.initializeUittreksels.perform();
+  }
+
+
+  @task
+  * initializeUittreksels() {
+    const uittreksels = [];
+    const prePublish = yield this.createPrePublishedResource.perform();
+    for(const uittreksel of prePublish) {
```
abeforgit, line 25: we could maybe extract this in a separate task and then run them concurrently with all() from ember-concurrency

PR #30, opened 2020-11-20
New draft decision statuses

Asergey91:


nvdk (changes requested):
I think this can be improved a bit, I'm missing some of the things we discussed:
- no search on the power select to select existing containers
- keeping document name and agendapoint name in sync

Other remarks:
- calling async functions in a component constructor is typically not a good idea or at least it requires a lot of extra checks in the async function. ember-concurrency has a good doc on the things you should do if you use a regular async function: http://ember-concurrency.com/docs/tutorial
- Please take the extra minute to properly name variables, using two letter variables is not ok in most cases (the exception perhaps the index in a for loop).
- In general we provide some space around assignments, so ``` const foo = "theValue"```

app/components/agenda/draft-import.js
```diff
@@ -2,24 +2,24 @@ import Component from '@glimmer/component';
...
-  async getDrafts(){
-    const containers=await this.store.query('document-container', {
+  @restartableTask
+  * getDrafts(){
+    const containers=yield this.store.query('document-container', {
       include: 'current-version,ontwerp-besluit-status',
       'filter[status][:id:]': 'c02542af-e6be-4cc6-be60-4530477109fc',
-      'filter[folder][:id:]': 'ae5feaed-7b70-4533-9417-10fbbc480a4c'
+      'filter[folder][:id:]': 'ae5feaed-7b70-4533-9417-10fbbc480a4c',
+      'page[size]': '1000'
```
nvdk, line 22: I'd rather have a proper search on the component instead of just requesting 1000 documents.

Asergey91: so not 100% sure what you mean. Do you want me to do a filter query onchange of the search bar in powerselect?

PR #35, opened 2020-12-07
Feature/demo bugfix

Asergey91:


app/components/agenda/modal.js
```diff
@@ -83,7 +83,7 @@ export default class AgendaModalComponent extends Component {
       this.toBeDeleted.push(this.currentlyEditing);
       this.isNew = false;
     } else {
-      this.currentlyEditing.rollbackAttributes();
+      // this.currentlyEditing.rollbackAttributes();
```
lagartoverde, line 86: As I said to arne, I think we should remove the commented code, in order to clean the repo :)

lagartoverde (approved):
Everything looks good and works 👍

app/components/agenda/modal.js
Asergey91: I am not sure about this line thats why I left it, rollBackAttributes behaviour is wierd when an empty agendapoint is created. but technically this should be here

app/components/agenda/modal.js
```diff
@@ -104,23 +104,27 @@ export default class AgendaModalComponent extends Component {
...
         const agendapoint = yield this.toBeDeleted[i];
         const behandeling = yield agendapoint.behandeling;
-        const documentContainer = yield behandeling.documentContainer;
-        if(documentContainer){
-          documentContainer.ontwerpBesluitStatus = this.store.findRecord('concept', 'a1974d071e6a47b69b85313ebdcef9f7'); //concept status
-          yield documentContainer.save();
+        if(behandeling){
+          const documentContainer = yield behandeling.documentContainer
+          if(documentContainer){
+            documentContainer.ontwerpBesluitStatus = this.store.findRecord('concept', 'a1974d071e6a47b69b85313ebdcef9f7'); //concept status
```
abeforgit, line 123: might be good to add a constant somewhere for the status ids

abeforgit (approved):

PR #36, opened 2020-12-07
Feature/optimize loading of participants

nvdk:
keep a store of potential participants on the top meeting-form component and
pass that down to where it's needed. I realized later this could also have been
a service, but I still think the approach works.

while doing this optimization I also refactored the loading of participants to
the treatment of an agendapoint, as this previously did not take pagination into
account. It now fetches the first 100 participants, which should suffice.


abeforgit (approved):
Fixed a small bug, looks good otherwise

nvdk:
much appreciated, thanks!

PR #37, opened 2020-12-07
Bugfix/voting

abeforgit:
Fixes display and persistence of non-voting participants to a vote


lagartoverde (approved):
Looks good for me.
I think is a little bit risky to include a new service last day before the demo, but if you trust it I trust you 👍 .
Also now the unanimous support button only works for the added voters, don't know if this is expected or a bug, but before it was quick to approve by unanimous support, just click on the button, now you have to click on each person to add them as a voter and then click the unanimous support button

PR #39, opened 2020-12-08
Feature/demo bugfix

Asergey91:
will still probably break in some edge cases like during race conditions


lagartoverde (approved):
Looks good an works for me 👍

PR #41, opened 2020-12-09
refactor: reuse meeting participation list component in treatment

nvdk:
this ensures the participants are shown in a consistent matter.


Asergey91:
There is some wierd behaviour on edit of treatment participants. When you update the list, save, and refresh edits dont get persisted. This was not the case before.

nvdk:
hmm that seems to be an issue with mu-cache. It seems the save of a behandeling currently triggers both a GET call for the relationship and a patch call at the same time. I'll see if that can be fixed easily

nvdk:
I believe this should be fixed in https://github.com/lblod/app-gelinkt-notuleren/commit/dbddb4b3dcf2c4d13c4e50a36666a15ab7abe161 . Can you confirm @Asergey91 ?

Asergey91:
sure, I'll take a look

PR #43, opened 2021-01-27
Bugfix/aanwezigen filter

nvdk:
addresses https://binnenland.atlassian.net/browse/GN-2046 . 
It seems we already had the correct data to filter in our store and that was even used in the old "aanwezigen" plugin, but I guess somehow got lost during conversion to a regular ember component. 

This code assumes we have a <http://mu.semte.ch/vocabularies/ext/hasDefaultType> relationship on the "bestuursorgaan-classification-code" to specify which "bestuursfunctie-rol-code" should be used to filter for the "aanwezigen" relationship.

NOTE: The code needs to run against the latest development branch of app-gelinkt-notuleren as it relies on some inverse relationships that were added

### DIFF
--- a/app/components/meeting-form.js
+++ b/app/components/meeting-form.js
@@ -69,9 +69,7 @@ export default class MeetingForm extends Component {
       page: { size: 100 } //arbitrary number, later we will make sure there is previous last. (also like this in the plugin)
     };
     const mandatees = yield this.store.query('mandataris', queryParams);
-    this.aanwezigenBijStart = Array.from(
-      mandatees.filter( (mandatee) => this.isValidMandateeForMeeting(mandatee) )
-    );
+    this.aanwezigenBijStart = mandatees;
   }
 
   @task
--- a/app/components/meeting-form.js
+++ b/app/components/meeting-form.js
@@ -28,8 +28,22 @@ export default class MeetingForm extends Component {
     this.zitting = this.args.zitting;
     this.loadData.perform();
   }
+
   get isComplete() {
     return this.loadData.lastSuccessful;
+  }
+
+  /**
+   * checks whether a mandatee can be a member of the meeting
+   * this means the meeting date should be between the mandatee's start and end
+   * the mandatee needs to have a "acting" (waarnemend) or "effective" (effectief) status
+   */
+  isValidMandateeForMeeting(mandatee) {
+    const startOfMeeting = this.zitting.gestartOpTijdstip ? this.zitting.gestartOpTijdstip : this.zitting.geplandeStart;
+    const hasValidStartDate = mandatee.start <= startOfMeeting;
+    const hasValidEndDate = isEmpty(mandatee.einde) || mandatee.einde > startOfMeeting;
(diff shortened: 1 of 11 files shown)


app/components/meeting-form.js
```diff
@@ -29,41 +33,50 @@ export default class MeetingForm extends Component {
...
+      include: 'is-bestuurlijke-alias-van,status',
       sort: 'is-bestuurlijke-alias-van.achternaam',
       'filter[aanwezig-bij-zitting][:id:]': this.zitting.get('id'),
       page: { size: 100 } //arbitrary number, later we will make sure there is previous last. (also like this in the plugin)
     };
     const mandatees = yield this.store.query('mandataris', queryParams);
-    this.aanwezigenBijStart = Array.from(mandatees.filter( (mandatee) => isEmpty(mandatee.einde) || mandatee.einde > new Date()));
+    this.aanwezigenBijStart = Array.from(
+      // TODO: should this not filter on meeting date?
+      mandatees.filter( (mandatee) => (isEmpty(mandatee.einde) || mandatee.einde > new Date()) && [statusEffectief, statusWaarnemend].includes(mandatee.get("status.id")))
```
lagartoverde, line 60: Maybe the filter is too complex, I would put it in different lines so we can understand it better

app/components/meeting-form.js
```diff
@@ -5,10 +5,14 @@ import { task } from "ember-concurrency-decorators";
 import { inject as service } from "@ember/service";
 import { isEmpty } from '@ember/utils';
 
+const statusEffectief = '21063a5b-912c-4241-841c-cc7fb3c73e75';
```
lagartoverde, line 8: Maybe we can add on a comment what do these statuses mean in English?

app/components/meeting-form.js
```diff
@@ -29,41 +33,50 @@ export default class MeetingForm extends Component {
...
       sort: 'is-bestuurlijke-alias-van.achternaam',
       'filter[bekleedt][bevat-in][:uri:]': this.bestuursorgaan.get('uri'),
+      'filter[bekleedt][bestuursfunctie][:id:]': stringifiedDefaultTypeIds,
       page: { size: 100 } //arbitrary number, later we will make sure there is previous last. (also like this in the plugin)
     };
     const mandatees = yield this.store.query('mandataris', queryParams);
-    this.possibleParticipants = Array.from(mandatees.filter( (mandatee) => isEmpty(mandatee.einde) || mandatee.einde > new Date()));
+    this.possibleParticipants = Array.from(
+      // TODO: should this not filter on meeting date?
+      mandatees.filter( (mandatee) => (isEmpty(mandatee.einde) || mandatee.einde > new Date()) && [statusEffectief, statusWaarnemend].includes(mandatee.get("status.id")))
```
lagartoverde, line 78: Same as the other filter

app/components/meeting-form.js
lagartoverde: Also the TODO is resolved?

lagartoverde (approved):
Looks good to me, some small improvements in the comments

PR #46, opened 2021-01-28
Feature/default participants on agenda

nvdk:
GN-1959
start out without any participants on treatments and have the modal preset with
the participants of the meeting. This allows us to present the latest state of
the participants of the meeting and a user only has to click save to persist
this state.

Key differences:

OLD: treatments could have a different set on participants if the participants
of the meeting where changed between creating two agendapoints.

NEW: the user will have to fill in participants for every treatment
NEW: the user is presented with the participants of the meeting when setting the
participants of a treatment for the first time


app/components/participation-list.js
```diff
@@ -9,9 +9,17 @@ import { isEmpty } from '@ember/utils';
...
   @tracked info;
+  @tracked chairman;
+  @tracked secretary;
 
   @service store;
 
+  constructor() {
+    super(...arguments);
+    this.chairman = this.args.chairman ? this.args.chairman : this.args.defaultChairman;
+    this.secretary = this.args.secretary ? this.args.secretary : this.args.defaultSecretary;
```
lagartoverde, line 20: I like the translation of the terms, thank you, it makes it easier to read 👍

lagartoverde (approved):
Seems perfect to me

PR #47, opened 2021-01-29
make sure we properly remove all new items on cancel

nvdk:


lagartoverde (approved):
Not much to say, good job :)

PR #49, opened 2021-02-03
If we call load agenda again it resets the list and remove the deleted ones

lagartoverde:


nvdk (changes requested):
I don't think this is the best solution to be honest, we should avoid expensive queries like this one as much as possible. Definitely so since we should already have all data necessary.  Is the agendapoints we receive as a parameter not up to date? If so we should address that.

lagartoverde:
Ook, so I found a better solution, the problem is that the object we were receiving in the agenda points was identified as the same one, if you only deleted an item, so the solution would be to refresh the UI (something we did with the loadAgenda call as it removed the agenda part and replaced by a loader) or to modify the value we set on the variable so it's identified as different, in this case I found that even if the agenda points variable is thought to be equal its objects property is not because it has one less item.
I don't like a lot the solution because is messing up with the ember internals, but I didn't found anything else

nvdk:
Can you check if using https://api.emberjs.com/ember/release/classes/EmberObject/methods/notifyPropertyChange?anchor=notifyPropertyChange works, that would be a lot cleaner. In the long(er) term the agenda component should use proper ember array as those have good change tracking.

PR #50, opened 2021-02-04
when using the queryNestedRelationship function we didn't allow to set the page size and it defaulted to 20

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-2060


nvdk (approved):
looking good

PR #52, opened 2021-02-08
Added ttl file containing new besluit types

lagartoverde:


nvdk:
at some point these really need to move out of the frontend :D

PR #53, opened 2021-02-11
Feature/rework publish flow

Dietr:
Converted the publish flow to Appuniversum components and updated the design.

Some issues that are not yet fixed:
- WuSwitch is still present as there is no Appuniversum component yet
- The switch functionality (public/private) also logs an error and breaks the publish

Tested with the meeting of Gemeente Brakel.


nvdk:
I could not reproduce the error on the switch (when publishing notulen?), so merging as is.

PR #54, opened 2021-02-12
Feature/environment name in title

lagartoverde:


nvdk:
@lagartoverde probably to only set it, if it's provided? i.e. I don't think we want to set it on production.

lagartoverde:
You are right sorry, I will change it, didn't think about that

PR #57, opened 2021-02-17
Persist absentees as wel as participants

abeforgit:
fixes https://binnenland.atlassian.net/browse/GN-1810
linked:
https://github.com/lblod/app-gelinkt-notuleren/pull/19
https://github.com/lblod/notulen-prepublish-service/pull/18

### DIFF
--- a/app/components/participation-list.js
+++ b/app/components/participation-list.js
@@ -51,12 +51,6 @@ export default class ParticipationListComponent extends Component {
     return sorted;
   }
   get mandateesNotPresent() {
-    // if(this.aanwezigenBijStart.length > 0 && this.possibleParticipants) {
-    //   const aanwezigenUris = this.aanwezigenBijStart.map((mandataris) => mandataris.uri);
-    //   const notPresent = this.possibleParticipants.filter((mandataris) => !aanwezigenUris.includes(mandataris.uri));
-    //   const sorted=notPresent.sortBy('isBestuurlijkeAliasVan.achternaam');
-    //   return sorted;
-    // }
     const sorted = this.afwezigenBijStart.sortBy("isBestuurlijkeAliasVan.achternaam");
     return sorted;
   }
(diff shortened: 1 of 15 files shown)


abeforgit:
still need to update the prepublisher

abeforgit:
also just realised I still need to do behandelingen, so disregard for now

app/components/participation-list.js
```diff
@@ -37,13 +51,14 @@ export default class ParticipationListComponent extends Component {
...
-      const sorted=notPresent.sortBy('isBestuurlijkeAliasVan.achternaam');
-      return sorted;
-    }
-    return [];
+    // if(this.aanwezigenBijStart.length > 0 && this.possibleParticipants) {
+    //   const aanwezigenUris = this.aanwezigenBijStart.map((mandataris) => mandataris.uri);
+    //   const notPresent = this.possibleParticipants.filter((mandataris) => !aanwezigenUris.includes(mandataris.uri));
+    //   const sorted=notPresent.sortBy('isBestuurlijkeAliasVan.achternaam');
+    //   return sorted;
+    // }
```
lagartoverde, line 59: Why we comment out this part of the code? If it's not needed I would remove it

app/components/participation-list/modal.js
```diff
@@ -1,20 +1,42 @@
 import Component from '@glimmer/component';
 import { action } from "@ember/object";
-import { tracked } from '@glimmer/tracking';
+import { tracked } from 'tracked-built-ins';
```
lagartoverde, line 3: What's the difference between glimmer tracking and tracked-built-ins?

lagartoverde (approved):
Looks good to me, just check the 2 comments I added :)

app/components/participation-list/modal.js
abeforgit: https://github.com/pzuraq/tracked-built-ins
Tracked-built-ins can track things like native Maps and Arrays
The tracked decorator they provide works like the normal decorator for normal values, but can then also be used to track the builtin objects

app/components/participation-list.js
abeforgit: Because I forgot :stuck_out_tongue:

PR #59, opened 2021-02-23
Changing last successful by is idle should solve all the problems

lagartoverde:
When changing to the publication page with a slow connection it shows the last published document for a few seconds (depends on how much time it takes to load the new prepublish document), if we try to publish this document errors happen.
By changing the .lastSuccessful attribute by .isIdle we solve this problem as we place the loading screen as soon as the user enters the screen

### DIFF
--- a/app/controllers/meetings/publish/agenda.js
+++ b/app/controllers/meetings/publish/agenda.js
@@ -44,6 +44,14 @@ export default class MeetingsPublishAgendaController extends Controller {
     this.spoedeisendeAgenda = yield this.initializeAgenda.perform("spoedeisendeagenda");
   }
 
+  @task
+  * reloadAgendas() {
+    this.ontwerpAgenda = yield this.initializeAgenda.perform("ontwerpagenda");
+    this.aanvullendeAgenda = yield this.initializeAgenda.perform("aanvullendeagenda");
+    this.spoedeisendeAgenda = yield this.initializeAgenda.perform("spoedeisendeagenda");
+  }
+
+
   /**
    * @this {MeetingsPublishAgendaController}
    * @param {string} type
@@ -89,7 +97,7 @@ export default class MeetingsPublishAgendaController extends Controller {
 
     const id = this.model.id;
     yield fetch(`/signing/agenda/sign/${agendaType}/${id}`, { method: 'POST'});
-    yield this.initializeAgendas.perform();
+    yield this.reloadAgendas.perform();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


lagartoverde:
I'm checking this because it seems that it mess up with some publications

nvdk:
I don't really get why we need to code duplication for the tasks, something to be discussed in some more detail I think

abeforgit:
I think you can get the same behavior with a tracked property
like 
```
@tracked hasInitialized;

@task
* reloadAgendas() {
//... do stuff
this.hasInitialized = true
}
```
and then use that property in the template

But I admit it's less intuitive than I first thought, there doesn't seem to be an ootb way to ask for "has succesfully ran at least once" in ember-concurrency

abeforgit:
oh this might be a nice case for an encapsulated task
http://ember-concurrency.com/docs/encapsulated-task

then you could do the same but set the hasInitialized property on the task object itself, calling  it hasSucceededOnce or something

abeforgit (approved):
seems to work, though I can't really test properly since I never saw the original problem

As mentioned above there are some opportunities to reduce duplication, but I'll leave it up to @nvdk whether you should spend time on them, I don't see the current code as a major problem since the whole app already suffers from tons of duplication in tasks anyway, and we'd need a good thinking session how to handle it

lagartoverde:
The problem is that the task status is saved between the different publication because the controller is a singleton, so the isInitialized property is going to be kept between publications and the same problem as lastSuccesfull will happen

nvdk:
isn't the proper solution to just reset the controller state when navigation to or away from the controller? ember has the necessary abstractions to do that...

nvdk:
Looking at it again, isn't the main thing that we should show the loading screen if the task  has the isRunning state?

lagartoverde:
Yes, we have to show the loading screen when the task has the isRunning state, but only the first time it loads, the other times it will be triggered by publish or sign, and the component itself handles the loading

lagartoverde:
I'm looking into resetting the task statuses, I found this but not sure if it's what we want
https://github.com/machty/ember-concurrency/pull/253

nvdk:
But that's just because check the wrong thing (this should have been a chat :) )

e.g. 

if it's loading show the loader
else if it's successfull show the preview
else show the error

this should always work

nvdk:
I still don't like it, but properly fixing this would require quite some rework of the publication components. So I agree this is currently the best approach to fixing this.

PR #61, opened 2021-03-01
Agendapoint participation list was not updating the secretaris and the chairman correctly

lagartoverde:


abeforgit (approved):
looks good

weird that we missed this for so long haha

PR #66, opened 2021-03-05
modernize editor-document-title component

nvdk:
refactored the component to a glimmer component, but also changed how it works
slightly:

- removed integration with the document-title plugin, as that plugin is no
longer in use
- if a document doesn't have a name yet, immediately start in edit mode
- provided a button to toggle edit mode as clicking on the h1 was no longer
possible
- the button has the added benefit of being more accessible

it was not my goal to change how this component worked, so this should not be
seen as a solution to other usability issues


abeforgit (approved):
Looks good

Not great UX wise cause the focus keeps being stolen at random times, but as you said that's not the goal of this pr

abeforgit:
wait I noticed some thins after I added my review

- the edit button does not have cursor: pointer set on it 
- when you click the edit button the input field is editable but not focused, which means you have to click the input field again

Dietr (commented):
Added some styling to make the editing of the title clearer. Also added a close button as the focus change was acting a bit strange and inconsistent.

I do have some trouble with closing the edit mode when adding a new agendapoint. With existing agendapoints everything works fine.

nvdk:
changed the behaviour a bit (closing edit mode without setting a title will now trigger an error class).

PR #70, opened 2021-03-22
Feature/print styles

Dietr:
Update of the print css and styling of the "uittreksel" print view.


nvdk:
checked print view in of edit view of agendapoint, meeting view and the extract print view, looks ok for me!

PR #71, opened 2021-03-23
Feature/disable published editing

Asergey91:
this is a very quick and dirty implementation of this feature, should be discussed with dieter probably


nvdk:
For the published agendapoints in the meeting overview I guess we should create read only mode for the participants and voting components. That could be as simple as just hiding the button if `@readOnly` is true?

lagartoverde:
How should I test this? I created an agenda point, I filled the behandeling, published the notulen and I can still edit it

lagartoverde (approved):
Works good for me :)

nvdk:
I suggest we only merge when this is user friendly enough so as not to block any release.

Dietr (approved):
Updated the styling:
- added an info message on the agendapoint and enabled the rdfa annotations (i guess we should still show that even when published)
- added a published label in the treatment and some minor cosmetic changes

Documented the current logic/design and an idea for a future enhancement here: https://abb-vlaanderen.gitbook.io/feature-passports-lblod/gn/agendapoints/published-meeting-agendapoint-in-development 
To be discussed, feel free to add your comments.

PR #78, opened 2021-04-14
Fix the remaining linting errors

Windvis:
~~Based on #77 since that one also fixed some errors. I can rebase if needed.~~ I dropped the commit from this PR so this can be merged first if needed.

### DIFF
--- a/app/components/agenda-manager/agenda-table/row.hbs
+++ b/app/components/agenda-manager/agenda-table/row.hbs
@@ -1,8 +1,8 @@
 <DraggableObject @tagName="tr" @content={{@item}} @isSortable={{true}}>
   <td>
-    <AuButton @skin="tertiary">
+    <AuButton @skin="tertiary" class="au-u-cursor-inherit">
       <span class="au-u-hidden-visually"></span>
-      <span style="cursor:pointer">
+      <span>
         <i class="fa fa-bars"></i>
       </span>
     </AuButton>
--- a/app/components/participation-list.hbs
+++ b/app/components/participation-list.hbs
@@ -38,19 +38,20 @@
   </ul>
 </div>
 {{/if}}
-{{#if (not @readOnly)}}
-<AuButton
-  @width="block"
-  @skin="secondary"
-  @icon="pencil"
-  @iconAlignment="left"
-  {{on 'click' this.togglePopup}}
-  @disabled={{this.fetchMandatees.isRunning}}
-  class="au-u-margin-top-small"
->
-  {{@modalTitle}}
-</AuButton>
-{{/if}}
+
+{{#unless @readOnly}}
+  <AuButton
+    @width="block"
+    @skin="secondary"
+    @icon="pencil"
+    @iconAlignment="left"
+    {{on 'click' this.togglePopup}}
+    @disabled={{this.fetchMandatees.isRunning}}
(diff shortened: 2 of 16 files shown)


app/components/agenda-manager/agenda-table/row.hbs
```diff
@@ -1,11 +1,6 @@
 <DraggableObject @tagName="tr" @content={{@item}} @isSortable={{true}}>
-  <td>
-    <AuButton @skin="tertiary">
```
Windvis, line 3: Button seemed unnecessary since the whole tr is draggable. It also added unwanted styles to the icon, which is why the inline styles were needed in the first place.

app/components/participation-list.hbs
```diff
@@ -38,19 +38,20 @@
...
-  @iconAlignment="left"
-  {{on 'click' this.togglePopup}}
-  @disabled={{this.fetchMandatees.isRunning}}
-  class="au-u-margin-top-small"
->
-  {{@modalTitle}}
-</AuButton>
-{{/if}}
+
+{{#unless @readOnly}}
```
Windvis, line 42: I'm really not a fan of this. I think the `{{#if (not something)}}` combo is way more readable, but the linting rule doesn't have an option to allow it.

app/controllers/meetings/publish/notulen.js
```diff
@@ -68,7 +69,14 @@ export default class MeetingsPublishNotulenController extends Controller {
     const id = this.model.id;
     const response = yield fetch(`/prepublish/behandelingen/${id}`);
     const json = yield response.json();
-    return json.map((response) => response.data.attributes);
+
+    return json.map((response) => response.data.attributes)
+            .map((behandeling) => {
+              return {
+                ...behandeling,
+                content: htmlSafe(behandeling.content),
```
Windvis, line 77: Maybe it would be better to install / write an html-safe helper so this can be done in the template where needed.

nvdk: quite sure we have a html-safe helper and use it elsewhere, so that's definitely an option :+1:

app/components/agenda-manager/agenda-table/row.hbs
nvdk: I think the button was added for accessibility reasons, but not 100% sure.

app/controllers/meetings/publish/notulen.js
Windvis: Woops, not sure how I missed that :smile:.

app/components/agenda-manager/agenda-table/row.hbs
Windvis: Yea, that makes sense. I don't think only adding the button will make it accessible though. It doesn't have any keyboard navigation as far as I can tell and no readable instructions either. But I'll add it back anyway.

Windvis: Reverted to the previous version and added the cursor override.

app/controllers/meetings/publish/notulen.js
Windvis: Updated.

nvdk (approved):
changes look sane and some basic testing did not reveal any bugs. ok to merge for me!

PR #79, opened 2021-04-15
Remove the ember-simple-auth mixins

Windvis:
This replaces the ember-simple-auth mixins with session service based alternatives.

### DIFF
--- a/app/routes/import.js
+++ b/app/routes/import.js
@@ -1,24 +1,16 @@
 import Route from '@ember/routing/route';
-import { computed } from '@ember/object';
-import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import { inject as service } from '@ember/service';
 
-export default Route.extend(AuthenticatedRouteMixin, {
-  queryParams: {
+export default class ImportRoute extends Route {
+  @service session;
+
+  queryParams = {
     mock: { refreshModel: true },
-    source: { refreshModel: true }
-  },
+    source: { refreshModel: true },
+  };
 
   beforeModel(transition) {
-    this.set('mock', transition.to.queryParams.mock);
-    this._super(...arguments);
-  },
-
-  authenticationRoute: computed('mock', function() {
-    if (this.mock) {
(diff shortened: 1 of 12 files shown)


app/routes/login.js
```diff
@@ -1,4 +1,10 @@
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
 
-export default Route.extend({
-});
+export default class LoginRoute extends Route {
+  @service session;
+
+  beforeModel() {
+    this.session.prohibitAuthentication('index');
```
Windvis, line 8: The previous version didn't redirect when the user was already logged in.

app/routes/import.js
```diff
@@ -1,24 +1,16 @@
...
-  authenticationRoute: computed('mock', function() {
-    if (this.mock) {
-      return 'mock-login';
-    }
-    else {
-      return 'login'; // TODO read from configuration
-    }
-  })
-});
+    let loginRoute = transition.to.queryParams.mock ? 'mock-login' : 'login';
```
Windvis, line 13: This is the only route that has this kind of logic, but I'm not sure if it is still needed?

nvdk: ah it does! this is actually a route other applications can redirect to if they want to pass information to GN and this way we can also transition to the correct route after login in testing scenario's.

Windvis: Aha ok! I'll leave it like this then.

nvdk (approved):
Looks to be working fine! I guess we can now also (re)enable the mixin linting rule :+1:

Windvis:
> Looks to be working fine! I guess we can now also (re)enable the mixin linting rule +1

@nvdk There are still some mu-semtech/ember-data-table mixins in use. Or do you mean enable it and add disable comments to those specific files?

nvdk:
right I keep forgetting about ember-data-table, we really should get around to modernizing that addon :). I guess we keep the rule as is for now, I don't see much benefit in disabling in the specific files as it's rather unlikely we'll introduce new mixins anyway.

PR #81, opened 2021-04-15
Modernize the CurrentSession service

Windvis:
This converts the CurrentSession service to a native class and removes all async properties since all they are guaranteed to be set once the application boots.

### DIFF
--- a/app/components/meeting-form.hbs
+++ b/app/components/meeting-form.hbs
@@ -4,7 +4,7 @@
       <AuLink @linkRoute="inbox.meetings" @skin="secondary" @icon="arrow-left" @iconAlignment="left">
         {{t "meetingForm.backButton"}}
       </AuLink>
-      <span class="au-c-app-chrome__entity">{{get (await this.currentSession.group) 'classificatie.label'}} {{get (await this.currentSession.group) 'naam'}}</span>
+      <span class="au-c-app-chrome__entity">{{this.currentSession.group.classificatie.label}} {{this.currentSession.group.naam}}</span>
     </AuToolbarGroup>
     <AuToolbarGroup>
       <ul class="au-c-list-horizontal au-u-padding-right-tiny">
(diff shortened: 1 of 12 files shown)


app/services/current-session.js
```diff
@@ -1,47 +1,42 @@
...
+
+  get canSign() {
+    return this.hasRole('GelinktNotuleren-ondertekenaar');
+  }
 
-export default Service.extend({
-  session: service('session'),
-  store: service('store'),
   async load() {
     if (this.session.isAuthenticated) {
```
Windvis, line 23: Should I prevent this from running more than once and show a warning if it does?

nvdk: I guess that could be a nice addition, but definitely not a requirement

app/components/meeting-form.hbs
```diff
@@ -4,7 +4,7 @@
       <AuLink @linkRoute="inbox.meetings" @skin="secondary" @icon="arrow-left" @iconAlignment="left">
         {{t "meetingForm.backButton"}}
       </AuLink>
-      <span class="au-c-app-chrome__entity">{{get (await this.currentSession.group) 'classificatie.label'}} {{get (await this.currentSession.group) 'naam'}}</span>
+      <span class="au-c-app-chrome__entity">{{this.currentSession.groupclassificatie.label}} {{this.currentSession.group.naam}}</span>
```
nvdk, line 7: missing a `.` here, should be `this.currentSession.group.classificatie.label`

nvdk (changes requested):
a small (but critical :) ) typo, but other than that looks ok!

app/components/meeting-form.hbs
Windvis: Woops, will fix!

Windvis: Should be good now.

app/services/current-session.js
Windvis: Yea it's probably overkill since that will be noticed during PR's anyway.

PR #83, opened 2021-04-16
Ember 3.24 LTS update

Windvis:
The Ember 3.24 blueprints included an updated linting config which caused a lot of errors :smile:. ~~Most of them related to prettier, so they were autofixable.~~ Disabled Prettier for now. There is still a sizeable amount that needs to be fixed manually though.

~~It's easier to check commit per commit since the prettier autofix touched _a lot_ of files.~~

~~I didn't fix the currentSession linting errors since those are fixed in #81. I'll rebase once that one is merged.~~ Done


.eslintrc.js
```diff
@@ -7,54 +7,54 @@ module.exports = {
...
   env: {
-    browser: true
+    browser: true,
   },
   rules: {
-    'ember/no-jquery': 'error',
-    "ember/no-mixins": "warn",
-    "semi": "error",
+    'ember/avoid-leaking-state-in-ember-objects': 'warn',
+    'ember/no-classic-classes': 'warn',
```
Windvis, line 27: Both of these rules can be removed once everything has been converted to native classes. I'll create a separate PR for that.

app/components/language-select.js
```diff
@@ -19,7 +21,7 @@ export default class LanguageSelect extends Component {
 
   @action
   selectLanguage(language) {
-    this.set('selected', language);
```
Windvis, line 22: This was a bug actually :smile:. Glimmer components don't have that method.

app/components/participation-list/mandatarissen-table.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable ember/no-empty-glimmer-component-classes */
```
Windvis, line 1: This file contains TypeScript JSDoc, so I kept it.

app/components/signatures/blockchain-status.js
```diff
@@ -1,6 +0,0 @@
-import Component from '@ember/component';
```
Windvis, line 1: The template file was [removed in a previous commit](https://github.com/lblod/frontend-gelinkt-notuleren/commit/ef9177a035f4ceaa88eacf3cca62c58290b375e3#diff-640d772fc45a12a869c2fdfaf71605911dd8ec442e266a1fb4c11b30db6969cb).

nvdk:
That is a big one indeed! I see prettier prefers single quotes, where I have a personal preference for double quotes... that will take some getting used to :D.

Windvis:
> That is a big one indeed! I see prettier prefers single quotes, where I have a personal preference for double quotes... that will take some getting used to :D.

@nvdk It's configurable though. This is just how it rolls out of the ember blueprint. So if it really bother you, it can be changed :smile:.

nvdk:
hmm I guess we will need to have a codestyle discussion after all, last time we tried that we didn't really reach a consensus :/. If possible, I think I'd prefer to ignore prettier rules for now so as not to block the upgrade until we've agreed upon codestyle rules.

Windvis:
> hmm I guess we will need to have a codestyle discussion after all, last time we tried that we didn't really reach a consensus :/. If possible, I think I'd prefer to ignore prettier rules for now so as not to block the upgrade until we've agreed upon codestyle rules.

Well that's the whole point of Prettier :smile: . You don't have to discuss anything, just use what it gives you, (and optionally configure some small subjective things like semicolons and quotes). 

I don't think disabling the rules is a good idea. The code wasn't consistent at all since the recommended Eslint config doesn't include any formatting related rules. It was one of the main reasons I wanted to do the update :smile:.

nvdk:
I'm not a fan of pushing these kinds of things through in a PR without a broader discussion, code style is a touchy subject for most developers and enforcing one is not something we did before.

abeforgit:
Big fan of prettier myself, I really dislike the endless (and imo pointless) discussions around code style
But yeah we have quite the spectrum of opinions on this topic so the discussions will have to happen anyway :sweat_smile:

Windvis:
> I'm not a fan of pushing these kinds of things through in a PR without a broader discussion

Well, this is just what the default blueprint configuration looks like in 3.24. It's not that I'm pushing my own preference here.

I know code style is a tricky subject, which is why I like that the Ember community decided for us by shipping Prettier in the blueprint. In theory that would have been one less thing to bike-shed / discuss / argue about 😄. I've had these kind of discussions before and it never ended in a decided upon code style.

So if there will be a discussion I really hope it's about which Prettier options we configure instead of deciding about using Prettier at all.

nvdk:
I think the discussion will mostly be:
- what tooling do we use for auto code formatting (with prettier being most likely)
- how do we configure said tool

It's a discussion we should have as it makes sense to have project and/or company wide defaults.

We knew prettier was coming to emberjs, I didn't realize they would push it so hard though. In any case I'd like to have the discussion before merging this PR.

Windvis:
I disabled the prettier rules for now. I can also split out the linting fixes into a separate PR if you prefer that, (and keep the update PR open until the decision has been made).

nvdk (changes requested):
It looks like the upgrade to 3.24 breaks our agenda-manager component, it is no longer correctly listing the agendapoint (on load, on refresh and after creating a new agendapoint). It could be this has to do with the use of tracked-built-ins.

PR #84, opened 2021-04-19
remove deprecated rdfa-editor overrides

Dietr:
Cleaned up the rdfa css styles to work with the latest version of ember-rdfa-editor (currently on the development branch).


nvdk (approved):
looks good to me, I'll merge when we bump ember-rdfa-editor

PR #85, opened 2021-04-19
Convert Ember Data Models to native classes

Windvis:
This converts all models that use the classic class syntax to native classes.

### DIFF
--- a/app/models/bestuursorgaan-classificatie-code.js
+++ b/app/models/bestuursorgaan-classificatie-code.js
@@ -1,14 +1,19 @@
 import Model, { attr, hasMany } from '@ember-data/model';
-import { collect } from '@ember/object/computed';
 
-export default Model.extend({
+export default class BestuursorgaanClassificatieCodeModel extends Model {
+  @attr label;
+  @attr scopeNote;
+  @attr uri;
+  @hasMany('bestuursfunctie-code', { inverse: 'standaardTypeVan' }) standaardType;
+  @hasMany('bestuursorgaan', { inverse: null }) isClassificatieVan;
+
   // A string representation of this model, based on its attributes.
   // This is what mu-cl-resources uses to search on, and how the model will be presented while editing relationships.
-  stringRep: collect.apply(this,['id', 'label', 'scopeNote']),
-
-  label: attr(),
-  scopeNote: attr(),
-  standaardType: hasMany('bestuursfunctie-code', { inverse: 'standaardTypeVan' }),
-  isClassificatieVan: hasMany('bestuursorgaan', { inverse: null}),
-  uri: attr()
-});
+  get stringRep() {
+    return [
+      this.id,
+      this.label,
+      this.scopeNote,
+    ];
+  }
+}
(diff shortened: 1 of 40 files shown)


app/models/bestuursorgaan-classificatie-code.js
```diff
@@ -1,14 +1,19 @@
...
   // This is what mu-cl-resources uses to search on, and how the model will be presented while editing relationships.
-  stringRep: collect.apply(this,['id', 'label', 'scopeNote']),
-
-  label: attr(),
-  scopeNote: attr(),
-  standaardType: hasMany('bestuursfunctie-code', { inverse: 'standaardTypeVan' }),
-  isClassificatieVan: hasMany('bestuursorgaan', { inverse: null}),
-  uri: attr()
-});
+  get stringRep() {
```
Windvis, line 12: I couldn't find any references in the codebase to this property. Is this still needed?

app/models/zitting.js
```diff
@@ -1,20 +1,23 @@
...
+  @hasMany('behandeling-van-agendapunt', { inverse: null }) behandeldeAgendapunten;
+  @belongsTo('editor-document', { inverse: null }) notulen;
+  @hasMany('agenda', { inverse: null }) publicatieAgendas;
+  @hasMany('agendapunt', { inverse: 'zitting' }) agendapunten;
+  @belongsTo('functionaris', { inverse: null }) secretaris;
+  @belongsTo('mandataris', { inverse: null }) voorzitter;
+  @hasMany('mandataris', { inverse: null }) aanwezigenBijStart;
+  @hasMany('mandataris', { inverse: null }) afwezigenBijStart;
+
+  get isPublished() {
```
Windvis, line 20: I don't think this is used anywhere?

app/models/kandidatenlijst.js
```diff
@@ -1,16 +1,17 @@
...
+  @attr lijstnummer;
+  @belongsTo('rechtstreekse-verkiezing', { inverse: 'heeftLijst' }) rechtstreekseVerkiezing;
+  @hasMany('persoon', { inverse: 'isKandidaatVoor' }) kandidaten;
+
+  rdfaBindings = {
     lijstnaam: "http://www.w3.org/2004/02/skos/core#prefLabel",
     lijstnummer: "http://data.vlaanderen.be/ns/mandaat#lijstnummer",
     kandidaten: "http://data.vlaanderen.be/ns/mandaat#heeftKandidaat",
     rechtstreekseVerkiezing: "http://data.vlaanderen.be/ns/mandaat#behoortTot",
     "lijsttype": "http://data.vlaanderen.be/ns/mandaat#lijsttype"
```
Windvis, line 15: The model doesn't have an attribute / relationship named like this, can this be removed?

app/models/bestuursorgaan-classificatie-code.js
nvdk: I believe this used to be used in an auto generated admin interface, where this probably originated. should be safe to remove now.

app/models/kandidatenlijst.js
nvdk: it seems like it does exist in the backend: https://github.com/lblod/app-gelinkt-notuleren/blob/f3b160e22bb63b239a728a2477c34411057f8d41/config/resources/slave-mandaat-domain.lisp#L47 so we either add the relationship or remove it from the bindings. both are fine for me

app/models/zitting.js
nvdk: I think this is related to old code indeed, should be safe to be removed.

nvdk (approved):
probably needs some testing to make sure, but code wise I don't see any issues,

app/models/kandidatenlijst.js
Windvis: I removed the binding since the `lijsttype` model doesn't exist yet in the frontend either.

PR #87, opened 2021-04-19
Feature/intermission

lagartoverde:


lagartoverde:
Please doublecheck the dutch translations there were some that weren't defined in the feature passport so maybe they are not correct. Also @Dietr wanted to tweak the design :)

Windvis:
@lagartoverde I think this should target the development branch. That branch has the updated CI config as well :smile:.

Windvis (changes requested):

app/components/manage-intermissions/delete.hbs
```diff
@@ -0,0 +1,43 @@
...
+      {{t "manageIntermissions.warning"}}
+    </AuHelpText>
+
+    {{#if this.isShowingWarning}}
+      <div id="isShowingWarning-verwijder-agendapunt">
+        <div class="au-u-margin-bottom">
+          <AuLabel>
+            {{t "manageIntermissions.agreeQuestion"}}
+          </AuLabel>
+          <label class="au-c-control au-c-control--checkbox">
```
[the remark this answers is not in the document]
lagartoverde: Basically I copied over the delete block of the agenda component, and that wasn't using it, I will change it

app/components/manage-intermissions/index.hbs
```diff
@@ -0,0 +1,59 @@
...
+          {{t "manageIntermissions.commentLabel"}}
+        </th>
+        <th class="au-u-table-right">
+          <span class="au-u-hidden-visually">{{t "manageIntermissions.actionsLabel"}}</span>
+          <AuButton
+            {{on "click" this.addIntermission}}
+            @icon="add"
+            @iconAlignment="left"
+          >
+            {{t "manageIntermissions.addInterruptionLabel"}}
```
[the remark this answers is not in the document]
lagartoverde: I think the passport description changed, In the current specification this is no longer the case, I talked with dieter and confirmed we have the correct behavior now :)

translations/en-us.yaml
```diff
@@ -258,3 +258,23 @@ blockchainStatus:
...
+  noDataMessage: No intermissions
+  modalTitle: Intermission
+  insertLabel: Insert
+  saveLabel: Save
+  cancelLabel: Cancel
+  editLabel: Edit Intermission
+  deleteCardTitle: Delete Intermission
+  warning: You cannot undo this action.
+  toggleWarningButton: Delete Intermission
+  agreeQuestion: Are you sure you want to delete the agenda item?
```
nvdk, line 277:
```suggestion
  agreeQuestion: Are you sure you want to delete the intermission?
```

translations/en-us.yaml
```diff
@@ -258,3 +258,23 @@ blockchainStatus:
...
+  checkboxLabel: I understand that the agenda item will be irreversibly removed.
```
nvdk, line 278:
```suggestion
  checkboxLabel: I understand that the intermission will be irreversibly removed.
```

nvdk (changes requested):
some small issues with the english translation.

@Dietr do we need the confirmation on removal do you think? I'm personally not a fan in the agenda manager and even less so here :)

app/components/manage-intermissions/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+  constructor() {
+    super(...arguments);
+    this.fetchIntermissions.perform();
+  }
+
+  @action
+  addIntermission() {
+    this.intermissionToEdit = this.store.createRecord("intermission", {
+      startedAt: new Date()
```
nvdk, line 21: nitpicking: I think it makes sense to default to the meeting startDate if that's available

Dietr:
> @Dietr do we need the confirmation on removal do you think? I'm personally not a fan in the agenda manager and even less so here :)

Not sure if we need it. For now I think it's best to keep it as is for consistency reasons. If we can rework the agendapoint delete at the same time I would prefer a simple delete button in the footer of the modal.

nvdk:
will merge once suggestiosn here and changes to backend have been merged, see https://github.com/lblod/notulen-prepublish-service/pull/20

PR #88, opened 2021-04-19
Convert the remaining classic classes to native classes

Windvis:
This builds on #83  since that has the updated linting config, but I'll rebase once reviewed.


app/adapters/application.js
```diff
@@ -1,29 +1,25 @@
...
 
-    const origAjax = this._super;
     const args = arguments;
     const sleep = (time) => new Promise( (resolve/*,reject*/) => setTimeout(() => resolve(true), time));
     const retry = (livesSpent) => {
       return new Promise((resolve, reject) => {
         sleep(250 * livesSpent).then(() => {
           if (livesSpent < maxRetries )
-            origAjax.apply(this,args).then( (res) => resolve(res), () => retry(livesSpent + 1).then(
+            super.ajax(...args).then((res) => resolve(res), () => retry(livesSpent + 1).then(
```
Windvis, line 14: I need to double check this tomorrow.

app/utils/promise-proxy-object.js
```diff
@@ -4,4 +4,4 @@ import PromiseProxyMixin from '@ember/object/promise-proxy-mixin';
 // see https://emberjs.com/api/ember/3.7/classes/PromiseProxyMixin for
 // info on how this is used.
 
-export default ObjectProxy.extend(PromiseProxyMixin);
+export default class PromiseProxyObject extends ObjectProxy.extend(PromiseProxyMixin) {}
```
Windvis, line 7: I think this util can be removed, the code that used this seems to be removed as well.

nvdk: should be ok to remove indeed :+1:

nvdk (changes requested):
looks ok, but indeed the ajax overwrite is now broken.

app/adapters/application.js
Windvis: Fixed the issue and rewrote the method while I was at it. I think it's more readable now (and easier to extract into a shareable util).

.eslintrc.js
```diff
@@ -26,8 +26,6 @@ module.exports = {
     browser: true,
   },
   rules: {
-    'ember/avoid-leaking-state-in-ember-objects': 'warn',
```
Windvis, line 29: These overrides are now no longer needed :tada:

PR #89, opened 2021-04-20
Replace <WuDatepicker> with <AuDatePicker>

Windvis:
This includes the ember-appuniversum update commit from #95 as well to make it easier to test.

### DIFF
--- a/app/components/date-time-picker.hbs
+++ b/app/components/date-time-picker.hbs
@@ -1,7 +1,7 @@
 <div class="au-o-grid au-o-grid--small">
   <div class="au-o-grid__item au-u-2-6">
     <AuLabel>Datum</AuLabel>
-    <WuDatepicker class="au-c-datepicker" @onChange={{this.onChangeDate}} @value={{this.date}}/>
+    <AuDatePicker @onChange={{this.onChangeDate}} @value={{this.date}} />
   </div>
   <div class="au-o-grid__item au-u-2-6 au-u-1-6@medium">
     <AuLabel>Uur</AuLabel>
--- a/app/components/date-time-picker.hbs
+++ b/app/components/date-time-picker.hbs
@@ -2,7 +2,12 @@
   <div class="au-o-grid__item au-u-2-6">
     {{#let (unique-id) as |id|}}
       <AuLabel for={{id}}>Datum</AuLabel>
-      <AuDatePicker @id={{id}} @onChange={{this.onChangeDate}} @value={{this.date}} />
+      <AuDatePicker
+        @id={{id}}
+        @onChange={{this.onChangeDate}}
+        @value={{this.date}}
+        @localization={{this.datePickerLocalization}}
+      />
     {{/let}}
   </div>
   <div class="au-o-grid__item au-u-2-6 au-u-1-6@medium">
--- a/app/components/date-time-picker.hbs
+++ b/app/components/date-time-picker.hbs
@@ -1,14 +1,20 @@
 <div class="au-o-grid au-o-grid--small">
   <div class="au-o-grid__item au-u-2-6">
-    <AuLabel>Datum</AuLabel>
-    <AuDatePicker @onChange={{this.onChangeDate}} @value={{this.date}} />
+    {{#let (unique-id) as |id|}}
+      <AuLabel for={{id}}>Datum</AuLabel>
+      <AuDatePicker @id={{id}} @onChange={{this.onChangeDate}} @value={{this.date}} />
+    {{/let}}
   </div>
   <div class="au-o-grid__item au-u-2-6 au-u-1-6@medium">
-    <AuLabel>Uur</AuLabel>
-    <Input class="au-c-input au-c-input--block" type='number' @input={{fn this.onChangeTime 'hours'}} @value={{this.hours}} max={{24}} />
+    {{#let (unique-id) as |id|}}
+      <AuLabel for={{id}}>Uur</AuLabel>
+      <Input class="au-c-input au-c-input--block" type='number' @input={{fn this.onChangeTime 'hours'}} @value={{this.hours}} max={{24}} id={{id}} />
+    {{/let}}
   </div>
   <div class="au-o-grid__item au-u-2-6 au-u-1-6@medium">
-    <AuLabel>Minuten</AuLabel>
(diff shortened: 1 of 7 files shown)


app/components/date-time-picker.js
```diff
@@ -1,10 +1,42 @@
 import Component from '@glimmer/component';
 import { action } from "@ember/object";
+import { guidFor } from '@ember/object/internals';
+
+const duetDatePickerLocalization = {
```
Windvis, line 5: Hardcoded it like this for now. Should this come from the .yml translation files?

app/components/date-time-picker.js
```diff
@@ -14,8 +46,20 @@ export default class DateTimePicker extends Component{
       this.minutes = this.date.getMinutes();
     }
   }
+
+  get isoDate() {
+    if (this.date) {
+      let isoString = this.date.toISOString();
+      return isoString.substr(0, isoString.indexOf('T'));
```
Windvis, line 53: AuDatePicker / <duet-date-picker> expects the value attribute as ISO-8601 without any time information. Ideally this shouldn't be needed since AuDatePicker can abstract that for us. I think passing the date instances makes more sense.

app/components/date-time-picker.js
```diff
@@ -14,8 +46,20 @@ export default class DateTimePicker extends Component{
...
+    }
+
+    return '';
+  }
+
   @action
-  onChangeDate(date) {
+  onChangeDate(isoDate) {
+    let date = new Date(isoDate);
```
Windvis, line 61: Ideally AuDatePicker would simply return a date / undefined / null.

app/styles/app.scss
```diff
@@ -56,7 +56,7 @@
 
 // WU COMPONENTS (DEPRECATED)
 @import 'ember-vo-webuniversum';
-@import 'ember-vo-webuniversum-data-table';
+// @import 'ember-vo-webuniversum-data-table'; // This imports the ember-data-table styles which make <duet-date-picker> not render correctly
```
Windvis, line 59: This conflicted with the styles of the date picker. I think it can just be removed though. Everything seems to be converted to AuDataTable already?

app/components/date-time-picker.js
nvdk: ah yes, that seems preferable (we use [ember-intl](https://github.com/ember-intl/ember-intl) which has a `t` helper).

app/components/date-time-picker.hbs
```diff
@@ -1,14 +1,25 @@
 <div class="au-o-grid au-o-grid--small">
   <div class="au-o-grid__item au-u-2-6">
-    <AuLabel>Datum</AuLabel>
-    <WuDatepicker class="au-c-datepicker" @onChange={{this.onChangeDate}} @value={{this.date}}/>
+    {{#let (concat this.uniqueId "-date") as |id|}}
+      <AuLabel for={{id}}>Datum</AuLabel>
+      <AuDatePicker
```
Windvis, line 5: Another thing I noticed is that typing in dates doesn't seems to work yet, while it does work in the <duet-date-picker> demo. Clearing the date input with backspace also throws an exception in the console since that isn't handled yet by AuDatePicker.

app/components/date-time-picker.hbs
```diff
@@ -1,14 +1,25 @@
 <div class="au-o-grid au-o-grid--small">
   <div class="au-o-grid__item au-u-2-6">
-    <AuLabel>Datum</AuLabel>
-    <WuDatepicker class="au-c-datepicker" @onChange={{this.onChangeDate}} @value={{this.date}}/>
+    {{#let (concat this.uniqueId "-date") as |id|}}
```
Windvis, line 3: I didn't want to add more unrelated code, but I think adding the [`unique-id` helper](https://emberjs.github.io/rfcs/0659-unique-id-helper.html) would be a better solution. There is a [polyfill](https://github.com/ctjhoa/ember-unique-id-helper-polyfill) for it, but [it's very easy to implement](https://github.com/ctjhoa/ember-unique-id-helper-polyfill/blob/master/addon/helpers/unique-id.js) as well so creating it in app is an option as well.

Windvis: Created a separate PR #90 for this.

nvdk: I've merged #90 :+1:

app/components/date-time-picker.hbs
nvdk: since this also did not work correctly in the old component, I'm not going to see that as blocking for now

nvdk: ah doing some more testing, it is a bit worse.

Windvis:
Blocked until I finish the Appuniversum work.

Windvis:
Reworked the PR based on the changes done in [ember-appuniversum](https://github.com/appuniversum/ember-appuniversum/pull/93/files). Once that one is merged / released this PR can be unblocked by bumping the version.

app/components/date-time-picker.js
```diff
@@ -14,16 +16,38 @@ export default class DateTimePicker extends Component{
       this.minutes = this.date.getMinutes();
     }
   }
+
+  get datePickerLocalization() {
+    return {
+      buttonLabel: this.intl.t('auDatePicker.buttonLabel'),
```
Windvis, line 22: I didn't include the `placeholder` since that needs to stay in sync with the adapter, and we don't override that one. Ideally the adapter would support the correct formatting based on the locale as well but I don't know how to implement that yet and it's probably not worth the time investment (yet).

app/components/date-time-picker.hbs
Windvis: Should be ok now.

app/components/date-time-picker.js
Windvis: Done. I decided to generate the day and month names since that's not something you want to manually configure for each language and the browser has those lists already.

nvdk (approved):

app/styles/app.scss
nvdk: @Dietr can you confirm we no longer need these styles? (I didn't see anything of so I assume they are no longer used)

Dietr: There was a small issue with the meeting tables but added a fix for that

Dietr (approved):
Works fine for me. I do see some design issues that we need to fix when using the datepicker in a modal because the new one doesn't render in a wormhole but we can improve on that later.

PR #90, opened 2021-04-20
Install the unique-id helper polyfill

Windvis:
More information: https://github.com/emberjs/rfcs/pull/659

This makes it easier to generate ids for input elements.


.template-lintrc.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
 module.exports = {
-  extends: 'octane'
+  extends: 'octane',
+  rules: {
+    'no-curly-component-invocation': {
```
Windvis, line 6: The rules don't detect `unique-id` as a built-in helper yet.

PR #91, opened 2021-04-23
Update to ember-power-select 4.1.4

Windvis:
This should also fix the deprecation warnings about `-in-element` when building the app. 

Ideally `@lblod/ember-rdfa-editor-besluit-type-plugin` should be updated as well because now there are still 2 versions of ember-power-select being installed. ember-cli-dependency-lint wouldn't like that :smile:.

### DIFF
--- a/app/components/agenda-manager/agenda-item-form/select-location.hbs
+++ b/app/components/agenda-manager/agenda-item-form/select-location.hbs
@@ -5,32 +5,32 @@
   <div class="au-o-grid au-o-grid--tiny">
     <div class="au-o-grid__item au-u-1-2">
       <PowerSelect
-              @class="select select--block"
-              @renderInPlace={{true}}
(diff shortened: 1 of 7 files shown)


app/components/agenda-manager/agenda-item-form/select-location.hbs
```diff
@@ -5,32 +5,32 @@
   <div class="au-o-grid au-o-grid--tiny">
     <div class="au-o-grid__item au-u-1-2">
       <PowerSelect
-              @class="select select--block"
-              @renderInPlace={{true}}
-              @placeholder={{t "manageAgendaZittingModalMove.positionSelectPlaceholder"}}
-              @options={{this.locationOptions}}
-              @verticalPosition="above"
-              @searchEnabled={{false}}
```
Windvis, line 13: This is now the default.

nvdk (approved):

PR #94, opened 2021-04-29
bugfix for when new aps were duplicated after setting them last in list

Asergey91:
solves GN-2581


nvdk:
@Asergey91 Indeed fixes the bug...  I've noticed there is a similar issues for "vooraan in de agenda" and "na agendapunt". Perhaps best to fix all of them in one go?

oh and next time, please take care to pick a proper description for both the commit and the PR ;)

Asergey91:
will do, sorry

PR #95, opened 2021-04-29
Update ember appuniversum

Windvis:
This version includes the updated datepicker and prevents AuModels `closeModal` from getting called twice.


nvdk:
Looks like 0.4.0 has a breaking change in the arguments of `AuLoader` (`@size` was replaced with `@padding`), which will need to changed when we bump ember-appuniversum

Windvis:
My bad, I thought this release only included the 2 PR's I made. Should have double checked.

Windvis:
That project really needs a changelog to be honest :smile:

nvdk (approved):

PR #96, opened 2021-04-29
Handle the case where the user clears the selected "draft"

Windvis:
Fixes GN-2350

### DIFF
--- a/app/components/agenda-manager/agenda-item-form/select-draft.js
+++ b/app/components/agenda-manager/agenda-item-form/select-draft.js
@@ -17,6 +17,10 @@ export default class AgendaManagerAgendaItemFormSelectDraftComponent extends Com
     this.getDrafts.perform();
   }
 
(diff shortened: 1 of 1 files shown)


app/components/agenda-manager/agenda-item-form/select-draft.js
```diff
@@ -17,6 +17,10 @@ export default class AgendaManagerAgendaItemFormSelectDraftComponent extends Com
     this.getDrafts.perform();
   }
 
+  get agendaItem() {
+    return this.args.model;
```
Windvis, line 21: I aliased it since I think using "model" is confusing, but it's just a quick solution, I think it would be better if it was just passed in with a better name to be honest.

nvdk: I'm definitely OK with renaming the argument!

PR #98, opened 2021-05-04
Read only role

Asergey91:
Solves GN-2536 and GN-2537

### DIFF
--- a/app/components/agenda-manager/agenda-table/body.hbs
+++ b/app/components/agenda-manager/agenda-table/body.hbs
@@ -1,9 +1,11 @@
 <SortableObjects @tagName={{"tbody"}}
                  @sortEndAction={{@onSort}}
                  @sortableObjectList={{@items}}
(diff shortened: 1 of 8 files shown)


Asergey91:
can both "GelinktNotuleren-schrijver" and "GelinktNotuleren-lezer" be true?

nvdk:
I think there was a miscommunication, I'd expect this PR to make the /inbox/agendapoints view to link to read only views of the agendapoint and to protect the /agendapoints/:id/edit view so you can't access it if you don't have the necessary rights.

nvdk (changes requested):
I don't think this PR quite covers the story in Jira. to be honest I think it doesn't do a lot well:
- The readonly agendapoints all list as published, even if they are not.
- You can still create agendapoints in read only mode.
- Nothing done on the /agendapoints route

I believe we discussed the scope of the story quite well during the planning session, please pay more attention next time.

Asergey91:
my bad, oscar will be taking over

PR #99, opened 2021-05-06
feature: free text at start/end of a meeting

Asergey91:
solves gn-2587


Windvis (changes requested):
I compared the current implementation with the passport and it seems that it doesn't really follow it at the moment

Things I've noticed:
- The input fields should be listed under a new section title (and have a sidebar link as well)
- The alert / help text is missing
- I don't think an accordion is needed
- The editor should be a rich text editor

tests/integration/components/zitting-tekst-test.js
```diff
@@ -0,0 +1,26 @@
...
+    assert.equal(this.element.textContent.trim(), '');
+
+    // Template block usage:
+    await render(hbs`
+      <ZittingTekst>
+        template block text
+      </ZittingTekst>
+    `);
+
+    assert.equal(this.element.textContent.trim(), 'template block text');
```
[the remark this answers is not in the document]
nvdk: indeed empty tests should be removed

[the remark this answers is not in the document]
Asergey91: we generally dont write tests for the ember app, however we have quite a few on the editor. But yes you are right, removed the tests

app/components/zitting-tekst.hbs
```diff
@@ -0,0 +1,15 @@
+{{#if (eq @position "before")}}
+  <AuAccordion @accordionIconOpen="nav-down" @accordionIconClosed="nav-right" @accordionButtonLabel={{t 'manageFreeText.before'}}>
+    <AuTextarea id="text" @value={{this.text}} @width="block" disabled={{this.saveText.isRunning}}></AuTextarea>
+    {{#unless @readOnly}}
+      <AuButton {{on "click" (perform this.saveText)}} @loading={{or this.saveText.isRunning @readOnly}} @width="block" @skin="secondary">{{t 'manageFreeText.save'}}</AuButton>
+    {{/unless}}
+  </AuAccordion>
+{{else}}
+  <AuAccordion @accordionIconOpen="nav-down" @accordionIconClosed="nav-right" @accordionButtonLabel={{t 'manageFreeText.after'}}>
+    <AuTextarea id="text" @value={{this.text}} @width="block" disabled={{or this.saveText.isRunning @readOnly}}></AuTextarea>
```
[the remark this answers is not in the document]
Asergey91: Oh cool, didn't know that, thanks. However not really relavant anymore

app/components/zitting-tekst.hbs
```diff
@@ -0,0 +1,15 @@
+{{#if (eq @position "before")}}
```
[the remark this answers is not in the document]
Asergey91: yep, my bad i renamed them and forgot to change it

Asergey91:
well yeah, wasnt aware there is a feature passport for this, so basically rewrote the thing. Thanks @Windvis for the review.

app/components/meeting-form.hbs
```diff
@@ -130,15 +130,16 @@
...
           {{!-- Agenda --}}
-          <AuHeading id="sectionThree" @level="2" @skin="3" class="au-c-onboarding-wrapper au-u-margin-top-huge">
+          <AuHeading id="sectionThree" @level="2" @skin="3" class="au-c-onboarding-wrapper au-u-margin-top-huge au-u-margin-bottom-small">
             {{t "meetingForm.thirdSectionTitle"}}
             <span class="au-c-onboarding">
               <AuIcon @icon="info-circle"/>
               {{t "meetingForm.thirdSectionNotFilledWarning"}}
             </span>
           </AuHeading>
+          <ZittingTekst @zitting={{@zitting}} @position={{"intro"}} @readOnly={{true}}/>
```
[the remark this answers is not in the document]
Asergey91: not qute sure how I would go about simplifying it though, would love more input, will sync in rocket chat

app/components/meeting-form.hbs
```diff
@@ -138,13 +148,19 @@
...
           </AuHeading>
           <AgendaManager @zittingId={{@zitting.id}} @onSave={{perform this.fetchTreatments}} />
-
-          {{!-- Treatment of agenda --}}
+          
+          {{!-- Start of meeting --}}
           <AuHeading id="sectionFour" @level="2" @skin="3" class="au-c-onboarding-wrapper au-u-margin-top-huge">
             {{t "meetingForm.fourthSectionTitle"}}
+          </AuHeading>
+          <ZittingTextDocumentContainer @zitting={{@zitting}} @readOnly={{false}} @type={{"ext:intro"}}/>
```
Windvis, line 156: Nitpicky, but string arguments don't need curlies: `@type="ext:intro"`

Windvis: And the readOnly is hardcoded again (and not used either it seems?) :smile:.

app/components/zitting-text-document-container.js
```diff
@@ -0,0 +1,58 @@
...
+    showInsertButton: true,
+    showRdfa: false,
+    showRdfaHighlight: false,
+    showRdfaHover: false
+  };
+
+  @task
+  *getText() {
+    const zitting = yield this.args.zitting;    
+    if(zitting && !zitting.intro){
```
Windvis, line 27: Instead of catching the case where a zitting record doesn't have a certain attribute value, I would add empty strings for these on record creation. An alternative could be to give them default values in the Model class definition.

app/components/zitting-text-document-container.js
```diff
@@ -0,0 +1,58 @@
...
+  editorOptions = {
+    showToggleRdfaAnnotations: false,
+    showInsertButton: true,
+    showRdfa: false,
+    showRdfaHighlight: false,
+    showRdfaHover: false
+  };
+
+  @task
+  *getText() {
```
Windvis, line 25: If you keep the logic in the component, I would convert this to a getter instead.

app/components/zitting-text-document-container.js
```diff
@@ -0,0 +1,58 @@
...
+      this.text = yield zitting.intro;
+    }
+    else if (this.type === 'ext:outro') {
+      this.text = yield zitting.outro;
+    }
+  }
+
+  @task
+  *saveText() {
+    const zitting = yield this.args.zitting;
```
Windvis, line 40: As discussed, yield isn't needed here.

app/components/meeting-form.hbs
Asergey91: Yeah I need to still implement that, not sure if oscars branch is finalized, wanna have it consistant with how he does it

app/components/zitting-text-document-container.hbs
```diff
@@ -0,0 +1,59 @@
+
+<AuPill @icon="info-circle" @iconAlignment="left" class="au-u-margin-top-small">
```
Windvis, line 2: I would use the same structure as [the other sections of the page](https://github.com/lblod/frontend-gelinkt-notuleren/blob/4a6784c5217f8de2fac8462ae1a3cd0f1481d615/app/components/meeting-form.hbs#L91). It looks a bit different now.

Windvis:
One last thing: I noticed that the browser console shows an error [from somewhere in the editor](https://github.com/lblod/ember-rdfa-editor/blob/2401c946e4290273e6364d1dff94aa1a7b9c9b4b/addon/utils/ce/raw-editor.ts#L120).

`Uncaught TypeError: this.model is undefined`

app/components/zitting-text-document-container.hbs
Asergey91: well I mean it is in the feature passport

Windvis: Yea but it doesn't look the same as the others on the same page. They use a different class / html structure, right? Not the `AuPill` component.

Windvis: Screenshot to make it clearer what I mean:

image

Windvis: The passport version looks more like the first one.

Asergey91: ah fair enough

app/components/meeting-form.hbs
nvdk: @Asergey91 fyi just merged oscars branch!

app/components/zitting-text-document-container.hbs
```diff
@@ -0,0 +1,58 @@
...
+                    {{html-safe this.editor.htmlContent}}
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    {{else}}
+      <div class="container-flex--contain" resource="#" vocab="http://data.vlaanderen.be/ns/besluit#" typeOf={{@type}} prefix="ext: http://mu.semte.ch/vocabularies/ext/">
```
nvdk, line 38: there's no need to include the predicate here, but if you want to it should be in the `property` attribute and not in `typeof` (it's a property, not a type)

Asergey91:
> One last thing: I noticed that the browser console shows an error [from somewhere in the editor](https://github.com/lblod/ember-rdfa-editor/blob/2401c946e4290273e6364d1dff94aa1a7b9c9b4b/addon/utils/ce/raw-editor.ts#L120).
> 
> `Uncaught TypeError: this.model is undefined`

this has been fixed in pr https://github.com/lblod/ember-rdfa-editor/pull/104

nvdk (approved):

PR #100, opened 2021-05-06
Feature/read only meeting

lagartoverde:


nvdk (changes requested):
looks good, just need to hide the "new meeting" button under inbox/meetings

PR #101, opened 2021-05-12
<DateTimePicker> usability improvements

Windvis:
This makes it possible for the time picker to "wrap around" when certain
values are reached.

For example when the user types in 60 in the minutes field, the hour
field will be incremented by 1.

### DIFF
--- a/app/components/date-time-picker.js
+++ b/app/components/date-time-picker.js
@@ -1,20 +1,27 @@
 import Component from '@glimmer/component';
 import { action } from "@ember/object";
 import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
 
 export default class DateTimePicker extends Component{
   @service intl;
-  date;
-  hours;
-  minutes;
+  @tracked date;
 
   constructor() {
     super(...arguments);
     if(this.args.value) {
       this.date = new Date(this.args.value);
-      this.hours = this.date.getHours();
-      this.minutes = this.date.getMinutes();
+    } else {
+      this.date = new Date();
     }
+  }
+
+  get hours() {
+    return this.date.getHours();
+  }
+
+  get minutes() {
+    return this.date.getMinutes();
   }
 
   get datePickerLocalization() {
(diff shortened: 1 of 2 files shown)


app/components/date-time-picker.js
```diff
@@ -1,22 +1,29 @@
...
       this.date = new Date(this.args.value);
-      this.hours = this.date.getHours();
-      this.minutes = this.date.getMinutes();
+    } else {
+      this.date = new Date();
     }
   }
 
+  get hours() {
+    return this.date.getHours();
```
Windvis, line 20: This uses (abuses?) the fact that getHours will change the date when the value goes "out of bounds". 

- 24 -> date + 1, hour to 0
- -1 -> date - 1, hour to 23

app/components/date-time-picker.js
```diff
@@ -38,29 +45,29 @@ export default class DateTimePicker extends Component{
...
     const value = event.target.value;
-    if(!this.date) {
-      this.date = new Date();
-    }
+    let newDate = new Date(this.date.getTime());
+
     if(type === 'hours') {
-      if(value < 0 || value > 24) return;
-      this.date.setHours(value);
+      newDate.setHours(value);
```
Windvis, line 65: This uses (abuses?) the fact that getHours will change the date when the value goes "out of bounds". 

- 24 -> date + 1, hour to 0
- -1 -> date - 1, hour to 23

It works and felt weird to implement manually, but it's not very obvious when reading the code that it happens.

app/components/date-time-picker.js
```diff
@@ -38,29 +45,29 @@ export default class DateTimePicker extends Component{
...
     } else {
-      if(value < 0 || value > 60) return;
-      this.date.setMinutes(value);
+      newDate.setMinutes(value);
```
Windvis, line 67: Same here.

app/components/date-time-picker.js
```diff
@@ -1,22 +1,29 @@
...
+  @tracked date;
 
   constructor() {
     super(...arguments);
     if(this.args.value) {
       this.date = new Date(this.args.value);
-      this.hours = this.date.getHours();
-      this.minutes = this.date.getMinutes();
+    } else {
+      this.date = new Date();
```
Windvis, line 15: The current version always creates a new date when one isn't passed in, so it will always default to the current time in that case. This is needed because the time inputs are now derived from the date.

Would it make sense to not set a default value and disable the time inputs instead? The user would need to select a date first if non is provided? I think that might make more sense UX wise.

nvdk:
any particular reason you've closed this @Windvis ?

Windvis:
@nvdk I'm not sure when I will continue working on it, and open PRs bother me :smile:. The branch is still there so can always reopen the PR later.

PR #102, opened 2021-05-12
Solved bug, query nested relationship was not working on editing items with over 20 items in relationships

lagartoverde:


lagartoverde:
I solved the problem but I don't really like the solution, the problem was that we were doing nested relationships but we were getting the first relationship ids from the actual object, that was paginated, so our queries were also paginated even if we set in the options that they shouldn't be, so I had to query the original object again at the first line and then query all the sub relationships as we did before.
Any ideas on what we could do different? It seems weird to refetch an object we already have, even if the relationships are paginated

lagartoverde:
For reference this is the bug we are trying to solve https://binnenland.atlassian.net/browse/GN-2249

app/services/store.js
```diff
@@ -43,21 +43,25 @@ export default class ExtendedStoreService extends Store {
       firstRelationship
     );
     const isMany = firstRelInfo.kind === "hasMany";
+    const firstRelationshipData = (await this.query(modelInstance.get('constructor.modelName'), {
```
abeforgit, line 46: I think maybe a cleaner way is to change the signature of this method and work with the id of the modelInstance directly
that means we can use a `findRecord` to get the instance, and then go from there
I think there's no way around the fact that we have to fetch the object (again), but if we can we should let ember-data decide how, it might do some clever caching? or something? who knows 

In any case it will remove the illusion that this method does not refetch the modelinstance

abeforgit:
**WARNING** base branch is master, make sure it's intentional

app/services/store.js
lagartoverde: Not sure I understand what would be different between findRecord and query, and if we get only the record id I think we cannot get the modelName, we should also ask for that right?

abeforgit: iirc ember data can be smarter about findRecord
with query it will always make the request 
but tbh it probably doesnt matter

and yes, youd need the modelname as well

abeforgit: so in short: I think its better to just move to asking for modelname and id, along with the relationship path
that way we are sure that we have the right instance and its more clear to the caller we will make a full backend request 

@nvdk any thoughts?

nvdk: I understand the question is: what makes the most sense as a method signature?

In that case I think passing modelname and id, instead of the modelinstance is indeed more clear as to what will actually happen.

abeforgit: yeah
whether we use findrecord or query then shouldn't really matter here I think

nvdk: @lagartoverde  will you take this up?

lagartoverde: Yes :), probably won't finish today as I have to change every call to this method

abeforgit: I don't think it's used very often

lagartoverde:
Sorry for the delay, now it works and it's much simpler and doesn't use the weird method we were discussing, this method was only used in the fetchVoters on the editStemming and with @abeforgit we concluded it wasn't needed

lagartoverde:
It should be even faster because it makes less queries than before

lagartoverde:
I don't have data in Brabant (the add mock data didn't really work in my computer) could anyone check how much time does it takes?

nvdk (approved):
tested, response time is similar to what we had before but now we load all mandatees... so I guess it's slightly faster :)

PR #103, opened 2021-05-12
Agendapoint record cleanup

Windvis:
This rolls back the changes to the `Agendapunt` and `BehandelingVanAgendapunt` records when closing the modal without saving. This also means the newly created records are deleted.

This should fix both GN-2604 and GN-2605.

### DIFF
--- a/app/components/agenda-manager/agenda-context.hbs
+++ b/app/components/agenda-manager/agenda-context.hbs
@@ -4,6 +4,6 @@
   deleteItemTask=this.deleteItemTask
   onSort=(perform this.onSortTask)
   saveItemTask=this.saveItemTask
-  createItemTask=this.createItemTask
+  createAgendaItem=this.createAgendaItem
   onCancel=(perform this.resetItemTask)
 )}}
--- a/app/components/agenda-manager/agenda-context.hbs
+++ b/app/components/agenda-manager/agenda-context.hbs
@@ -1,8 +1,9 @@
 {{yield (hash
-                items=this.items
-                loadItemsTask=this.loadItemsTask
-                deleteItemTask=this.deleteItemTask
-                onSort=(perform this.onSortTask)
-                saveItemTask=this.saveItemTask
-                createItemTask=this.createItemTask
-        )}}
+  items=this.items
+  loadItemsTask=this.loadItemsTask
+  deleteItemTask=this.deleteItemTask
+  onSort=(perform this.onSortTask)
+  saveItemTask=this.saveItemTask
+  createItemTask=this.createItemTask
+  onCancel=(perform this.resetItemTask)
+)}}
--- a/app/components/agenda-manager/index.hbs
+++ b/app/components/agenda-manager/index.hbs
@@ -1,5 +1,6 @@
 <AgendaManager::AgendaContext
   @zittingId={{@zittingId}}
+  @onCreate={{this.editItem}}
   @onSave={{@onSave}}
   @onCancel={{this.stopEditing}}
   as |agenda|
@@ -11,7 +12,7 @@
     {{else}}
       <AgendaManager::AgendaTable
               @onSort={{agenda.onSort}}
-              @onCreate={{fn this.createAgendaItem agenda.createAgendaItem}}
+              @onCreate={{agenda.createAgendaItem}}
               @items={{agenda.items}}
               @readOnly={{@readOnly}}
               as |Table|>
--- a/app/components/agenda-manager/index.hbs
+++ b/app/components/agenda-manager/index.hbs
@@ -1,4 +1,9 @@
-<AgendaManager::AgendaContext @zittingId={{@zittingId}} @onSave={{@onSave}} as |agenda|>
+<AgendaManager::AgendaContext
+  @zittingId={{@zittingId}}
+  @onSave={{@onSave}}
+  @onCancel={{this.cancelEdit}}
+  as |agenda|
+>
   <div class="au-c-meeting-chrome-card">
     {{#if agenda.loadItemsTask.isRunning}}
       <AuLoader />
@@ -24,15 +29,16 @@
       </AgendaManager::AgendaTable>
 
     {{/if}}
-
   </div>
-  <AgendaManager::Edit @visible={{this.editModalVisible}}
-                       @itemToEdit={{this.itemToEdit}}
-                       @saveTask={{agenda.saveItemTask}}
-                       @onClose={{this.closeModal}}
-                       @deleteTask={{agenda.deleteItemTask}}
-                       @onCancel={{this.closeModal}}
-                       @agendaItems={{agenda.items}}
+
+  <AgendaManager::Edit
+    @visible={{this.editModalVisible}}
+    @itemToEdit={{this.itemToEdit}}
+    @saveTask={{agenda.saveItemTask}}
+    @onClose={{this.closeModal}}
+    @deleteTask={{agenda.deleteItemTask}}
+    @onCancel={{fn agenda.onCancel this.itemToEdit}}
+    @agendaItems={{agenda.items}}
   />
 
 </AgendaManager::AgendaContext>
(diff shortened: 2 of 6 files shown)


app/components/agenda-manager/agenda-context.hbs
```diff
@@ -1,8 +1,9 @@
 {{yield (hash
```
Windvis, line 1: Unrelated to these issues but I wanted to get the discussion started:

Personally I don't really see a benefit of the AgendaContext component. I understand the intention but in this case it only seems to add unnecessary indirection when trying to understand how the "AgendaManager" works. I think extracting that logic back into the AgendaManager component itself makes the most sense and would make things simpler and easier to follow.

app/components/agenda-manager/index.hbs
```diff
@@ -24,15 +29,18 @@
...
-  <AgendaManager::Edit @visible={{this.editModalVisible}}
-                       @itemToEdit={{this.itemToEdit}}
-                       @saveTask={{agenda.saveItemTask}}
-                       @onClose={{this.closeModal}}
-                       @deleteTask={{agenda.deleteItemTask}}
-                       @onCancel={{this.closeModal}}
-                       @agendaItems={{agenda.items}}
-  />
+
+  {{#if this.editModalVisible}}
```
Windvis, line 34: I prefer not rendering components at all when they aren't needed / used. (This also means the VM has less work to do, so it should have a slight impact on rendering performance.)

app/components/agenda-manager/index.js
```diff
@@ -13,35 +11,20 @@ import {task} from "ember-concurrency-decorators";
 
 /** @extends {Component<Args>} */
 export default class AgendaManagerIndexComponent extends Component {
-  /** @type AgendaData */
-  @tracked popup = false;
-  @tracked editModalVisible = false;
```
Windvis, line 18: Deriving the state from the `itemToEdit` prop seemed more logical since the edit modal will never be used if that prop is empty. Keeping a separate flag and keeping them in sync seemed unnecessary.

app/components/agenda-manager/index.js
```diff
@@ -13,35 +11,20 @@ import {task} from "ember-concurrency-decorators";
...
   }
 
   @task
   * createItemTask(newItemTask) {
     const newItem = yield newItemTask.perform();
     this.editItem(newItem);
   }
+
   @action
   cancelEdit() {
```
Windvis, line 27: `cancelEdit` isn't the best name anymore since it's now also used for simply closing the modal. `stopEditing` might be clearer?

nvdk: I think on closing the modal, the expectation is to also rollback changes. at least that's how the other modals work afaik. (in essence the close button should behave as a second cancel button).

app/components/agenda-manager/index.js
```diff
@@ -13,35 +10,20 @@ import {task} from "ember-concurrency-decorators";
...
 export default class AgendaManagerIndexComponent extends Component {
-  /** @type AgendaData */
-  @tracked popup = false;
-  @tracked editModalVisible = false;
   @tracked itemToEdit = null;
 
-  constructor(...args) {
-    super(...args);
+  get editModalVisible() {
+    return Boolean(this.itemToEdit);
```
abeforgit, line 16: I slightly prefer the double negative (as stupid as that syntax is) cause of the `new Boolean()` trap
but Im more than ok with leaving it as is

app/components/agenda-manager/index.js
abeforgit: makes sense

abeforgit (approved):
code looks good, unable to test at the moment

app/components/agenda-manager/agenda-context.hbs
abeforgit: I guess the hope was that you don't need to understand how the context methods work and just use them as a blackbox api
but I don't think I quite managed that with this code

app/components/agenda-manager/index.js
Windvis: I'm unfamiliar with the `new Boolean()` trap, can you explain / link me something? :smile: 

I just go for this version because I think it's more readable, but I don't mind switching either if there are downsides to this.

app/components/agenda-manager/index.js
Windvis: I think that's how it works now? I'll have to double check how I implemented it though. I have the memory of a fish :smile:.

app/components/agenda-manager/agenda-context.hbs
Windvis: Yea I understand your reasoning. I like that pattern as well, especially if it needs to be reusable. [I've used it recently myself](https://github.com/lblod/ember-rdfa-helpers/pull/8/files#diff-179a4536ca2838b76d93bda0f10394dd9c19f29ff2332505904c3c7adf6ccbcdR1-R8) to refactor away from a mixin.

Now that I look at it again, I think the thing I dislike about the current implementation is that it passes the `createAgendaItem` task as a param to the `createItemTask` on the parent component. It's harder to follow how everything works imo. The thing I like more is to pass the parent task as an argument to the context component, which can tell call it when needed (which is what I did in this PR for the reset functionality). I think that flow is easier to follow and more in line with how you interact with other components.

Anyway, it probably won't be worth the energy to remove it or change it since everything works already and it's just my opinion. So ignore my ramblings. :smile:

app/components/agenda-manager/index.js
Windvis: Just verified. `cancelEdit` is passed around to the nested components and used after other actions are complete:
- after saving
- after deletion
- after resetting

The reset functionality gets called every time the modal needs to be closed without using the save or delete buttons.

It's only used to close the modal, which is why the name isn't the best at the moment. I can just rename it to `closeModal` maybe, but it's a bit weird since the code in the action only clears the `itemToEdit` prop (which is why I suggested `stopEditing` I think).

nvdk:
Understood :). In that case stopEditing would make more sense indeed.

app/components/agenda-manager/index.js
Windvis: Renamed to `stopEditing`

app/components/agenda-manager/index.js
abeforgit: try this in your console:
```js
let b = new Boolean(false);
b ? "true" : "false";
```

Basically using the Boolean constructor can lead to confusing results. You're using it as a function here, which does what you expect. But coming from java or other languages seeing a function call with a PascalCase name can trigger a "huh I guess they forgot `new` here"

[more details](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean)

app/components/agenda-manager/agenda-context.hbs
abeforgit: ah yes so something like an onCreate argument on the context rather than the other way round? I agree that makes more sense, I wasn't happy with my solution either I recall

app/components/agenda-manager/index.js
Windvis: Ah yea. I see what you mean. Well, there are always code reviews to catch that, and an [eslint rule](https://eslint.org/docs/rules/no-new-wrappers) as well if things are _really_ getting out of hand :smile:.

I still prefer the `Boolean` version myself, since I think it's more readable. the `!!` version is also harder to google for. At least it was when I encountered it for the first time a couple of years ago :smile:.

Anyway, I think this is one of those things that should just be done consistently across a codebase so I'll change it.

Windvis: Changed to `!!`

app/components/agenda-manager/agenda-context.hbs
Windvis: I added the change since it was a small fix anyway.

app/components/agenda-manager/index.js
abeforgit: oh yea didnt know eslint would catch that
all good then

Windvis: It doesn't by default though. That rule isn't part of the recommended ruleset.

PR #104, opened 2021-05-12
Don't link to the "meeting" page after removing the agenda item

Windvis:
Fixes GN-2639 with a workaround, as discussed.

### DIFF
--- a/app/components/zitting-link.js
+++ b/app/components/zitting-link.js
@@ -17,7 +17,10 @@ export default class ZittingLinkComponent extends Component {
   @restartableTask
   * getMeeting(){
     const result = yield this.store.query("zitting", {
-      'filter[agendapunten][behandeling][document-container][:id:]':this.args.documentContainer.id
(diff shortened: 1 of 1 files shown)


app/components/zitting-link.js
```diff
@@ -17,7 +17,10 @@ export default class ZittingLinkComponent extends Component {
   @restartableTask
   * getMeeting(){
     const result = yield this.store.query("zitting", {
-      'filter[agendapunten][behandeling][document-container][:id:]':this.args.documentContainer.id
+      'filter[agendapunten][behandeling][document-container][:id:]': this.args.documentContainer.id,
+      // Including the relationships ensures the cache returns the proper response.
+      // TODO: This is a workaround for a mu-cache issue. Remove the include once that's resolved.
+      'include': 'agendapunten.behandeling.document-container'
```
nvdk, line 23: I think/hope that perhaps just including agendapunten may be enough (and will result in a smaller payload) :pray:

Windvis: Yup, that seems to work as well! I misunderstood and assumed the whole tree was needed.

PR #105, opened 2021-05-17
Feature: You can know search the chairman on firstname and lastname

lagartoverde:


nvdk:
probably best to also update the placeholder text, which currently states you can only search on last name

PR #106, opened 2021-05-17
Use auxiliar set to solve deprecation warning in participation list modal

lagartoverde:


Windvis:
It works, but I'm not a fan of the solution to be honest. I know I linked to an addon that solved it this way, but I saw it more of a last resort, if nothing else worked :smile:. Most of the time when I encountered this error it was possible to just restructure the code instead.

I would try one of the following:
1. convert `selectedMandatees` to a getter. On first glance it generates a map based on some of the arguments so I think that should work.
2. Treat the `selectedMandatees` map as immutable, create a local map in the method and assign it to the class property at the end

I think both of these should work (with the getter being my preference) and are less of a hack. I'm not 100% familiar with that part of the app so maybe I missing some details which makes this impossible though.

lagartoverde:
Yes, converting to a getter worked, I didn't fully understand what was the problem with the deprecation so I just followed the solution in the PR

app/components/participation-list/modal.js
```diff
@@ -39,6 +39,10 @@ export default class ParticipationListModalComponent extends Component {
     this.generateSelected();
   }
 
+  get selectedMandatees() {
+    return this.selectedMandateesMap;
```
Windvis, line 43: Hmm, that's not really what I meant with "convert selectedMandatees to a getter". The only reason this doesn't show the deprecation is [because it's not using a tracked map anymore](https://github.com/lblod/frontend-gelinkt-notuleren/pull/106/files#diff-f486ab0c87e76622c197ad111f9c13eea70c89eeea697ecf3065fbb15104b10eR33). If the map didn't need to be tracked in any way, just simply changing it like this would be enough to remove the deprecation:

`selectedMandatees = tracked(Map);`
to
`selectedMandatees = new Map();`

That's essentially what your new version does, only with an unneeded getter layer added on top.

What I meant was remove the selectedMandatees field on the class. Rename the `generateSelected` method selectedMandatees and convert it to a getter. Let that method create a local map variable and return that at the end. It would recompute if one of the arguments changes as well.

But like I said, if tracking isn't needed, simply removing tracked from the selectedMandatees map would be enough to stop the deprecation from showing.

Windvis (changes requested):

app/components/participation-list/modal.js
lagartoverde: I tested and ir worked without problems, I will retest because the tracking should be needed

lagartoverde: Yeah, tracking it's not needed, so much simpler

Windvis (approved):

PR #107, opened 2021-05-17
Replaced complex delete pattern by a simple button

lagartoverde:
Mentioning @Dietr so he can take a look at the styling and make the needed changes


Dietr (approved):

app/components/agenda-manager/edit.hbs
```diff
@@ -53,20 +53,30 @@
...
     <Modal.Footer>
+      <AuToolbar>
       <AuButtonGroup>
         <Form.Submit>{{t "manageAgendaZittingModal.saveButton"}}</Form.Submit>
         <AuButton @skin="secondary" {{on "click" this.cancel}}>
           {{t "manageAgendaZittingModal.cancelButton"}}
         </AuButton>
       </AuButtonGroup>
+      {{#unless this.isNew}}
+        {{!-- TODO: don't allow delete on published agendapoints  --}}
```
nvdk, line 68: since this is a pretty simple check to add (and deleting published agendapoints wasn't possible before), why not just add it?

lagartoverde: We can't edit published agenda points so we cannot delete them, I just tested it, I will delete the comment

PR #109, opened 2021-05-21
added a default value for free text in the model for sake of old meetings being backwards compatible

Asergey91:
very samll bugfix


Windvis (approved):
I assume mu-cl-resources doesn't have a way to set "default values"? I think that would be a better fix instead of catching that on the frontend. But this looks good to me otherwise!

PR #110, opened 2021-05-24
include calculated hash code when showing signature

nvdk:
depends on https://github.com/lblod/notulen-prepublish-service/pull/26


Asergey91:
nitpicky but formating is kinda broken, but I guess this is for dieter to look at

Asergey91 (approved):

nvdk:
> nitpicky but formating sometimes brakes the signature into 2 lines and it also kinda blends in with the timestamp, but I guess this is for dieter to look at

fair point, better ask @Dietr for some input :D

PR #115, opened 2021-05-26
Fix `<EditorDocumentTitle>` exception

Windvis:
I noticed that typing in a title showed an exception in the console.

Did a little refactoring while I was at it :smile:.

### DIFF
--- a/app/components/editor-document-title.hbs
+++ b/app/components/editor-document-title.hbs
@@ -5,7 +5,7 @@
 {{else}}
   {{#if this.active}}
     <div class="au-c-app-chrome__title-group {{if this.error "au-c-input--error"}}">
-      <Input class="au-c-app-chrome__title-input" placeholder="Naamloos document" @type="text" @value={{this.title}} {{on "input" this.setTitle}} @autofocus="autofocus"  />
+      <input class="au-c-app-chrome__title-input" placeholder="Naamloos document" type="text" value={{this.title}} {{on "input" this.setTitle}} />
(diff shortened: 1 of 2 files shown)


app/components/editor-document-title.hbs
```diff
@@ -5,7 +5,7 @@
 {{else}}
   {{#if this.active}}
     <div class="au-c-app-chrome__title-group {{if this.error "au-c-input--error"}}">
-      <Input class="au-c-app-chrome__title-input" placeholder="Naamloos document" @type="text" @value={{this.title}} {{on "input" this.setTitle}} @autofocus="autofocus"  />
```
Windvis, line 8: I removed the autofocus part since it seemed that it wasn't working. I think something like an `{{autofocus}}` modifier is needed.

app/components/editor-document-title.js
```diff
@@ -13,24 +12,16 @@ export default class EditorDocumentTitleComponent extends Component {
...
   @action
   setTitle(event) {
-    this._title = event.target.value;
-    if (this.args.onChange) {
-      this.args.onChange(this._title);
-    }
-    if (this._title) {
+    let title = event.target.value;
+
+    this.args.onChange?.(title);
```
abeforgit, line 22: wait you can optional chain a function call :open_mouth: 
mind = blown

abeforgit (approved):

app/components/editor-document-title.js
Windvis: Yea, it look's a bit weird at first but I highly prefer it to the alternatives now :smile:.

PR #117, opened 2021-05-28
Meeting creation flow changes

Windvis:
This changes the meeting creation flow. Instead of showing a "faux" edit page on which the user needs to click a button to open a modal, it now opens a modal directly. It also validates that an administrative body is selected before a meeting is created. On the edit page the administrative body is now readonly (GN-2651).

### DIFF
--- a/app/components/administrative-body-select.hbs
+++ b/app/components/administrative-body-select.hbs
@@ -1,0 +1,18 @@
+<div class={{if @error "ember-power-select--error"}}>
+  <PowerSelect
+    @placeholder={{t "manageZittingsData.selectPlaceholder"}}
+    @selected={{@selected}}
+    @options={{this.administrativeBodyOptions}}
+    @onChange={{@onChange}}
+    @triggerId={{@id}}
+    as |administrativeBody|
+  >
+    {{administrativeBody.isTijdsspecialisatieVan.naam}}
+    (periode: {{moment-format administrativeBody.bindingStart "MM/DD/YYYY"}} -
+    {{#if administrativeBody.bindingEinde~}}
+      {{moment-format administrativeBody.bindingEinde "MM/DD/YYYY"}}
+    {{~else~}}
+      nvt.
+    {{~/if}})
+  </PowerSelect>
+</div>
--- a/app/components/administrative-body-select.js
+++ b/app/components/administrative-body-select.js
@@ -1,0 +1,53 @@
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
+
+const VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS = [
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000005", //	"Gemeenteraad"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000007", //	"Raad voor Maatschappelijk Welzijn"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000009", //	"Bijzonder Comité voor de Sociale Dienst"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000a", //	"Districtsraad"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000c", //	"Provincieraad"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/53c0d8cd-f3a2-411d-bece-4bd83ae2bbc9", //	"Voorzitter van het Bijzonder Comité voor de Sociale Dienst"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/9314533e-891f-4d84-a492-0338af104065", //	"Districtsburgemeester"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000b", //	"Districtscollege"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/180a2fba-6ca9-4766-9b94-82006bb9c709", //	"Gouverneur"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/e14fe683-e061-44a2-b7c8-e10cab4e6ed9", //	"Voorzitter van de Raad voor Maatschappelijk Welzijn"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000006", //	"College van Burgemeester en Schepenen"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/4c38734d-2cc1-4d33-b792-0bd493ae9fc2", //	"Voorzitter van de Gemeenteraad"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e00000d", //	"Deputatie"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/4955bd72cd0e4eb895fdbfab08da0284", //	"Burgemeester"
+  "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000008" //	"Vast Bureau"
+];
+
+export default class AdministrativeBodySelectComponent extends Component {
+  @service currentSession;
+  @service store;
+  @tracked administrativeBodyOptions = [];
+
+  constructor() {
+    super(...arguments);
+
+    this.fetchAdministrativeBodies.perform();
+  }
+
+  @task
+  * fetchAdministrativeBodies() {
+    let currentAdministrativeUnitId = this.currentSession.group.id;
+
+    let administrativeBodiesInTime = yield this.store.query('bestuursorgaan', {
+      'filter[is-tijdsspecialisatie-van][bestuurseenheid][id]': currentAdministrativeUnitId,
+      'include': [
+        'is-tijdsspecialisatie-van.bestuurseenheid',
+        'is-tijdsspecialisatie-van.classificatie',
+      ].join(),
+      'sort': '-binding-start'
+    });
+
+    this.administrativeBodyOptions = administrativeBodiesInTime.filter((administrativeBodyInTime) => {
+      let classificationUrl = administrativeBodyInTime.get('isTijdsspecialisatieVan.classificatie.uri');
+      return VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS.includes(classificationUrl);
+    });
+  }
+}
--- a/app/components/meeting-form.hbs
+++ b/app/components/meeting-form.hbs
@@ -105,7 +105,6 @@
       </AuHeading>
       <Zitting::ManageZittingsdata
         @zitting={{@zitting}}
-        @onCreate={{@onCreateMeeting}}
         @onChange={{this.meetingInfoUpdate}}
         @readOnly={{this.readOnly}}
       />
--- a/app/components/meeting-form.hbs
+++ b/app/components/meeting-form.hbs
@@ -92,7 +92,10 @@
   <div class="au-c-meeting-chrome">
     <div class="au-c-meeting-chrome__paper">
       {{!-- Meeting title --}}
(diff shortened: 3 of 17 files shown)


app/templates/inbox/meetings.hbs
```diff
@@ -1,5 +1,5 @@
-<AuDataTable @content={{this.model}} @isLoading={{this.isLoadingModel}} @noDataMessage="Geen zittingen" @sort={{this.sort}} @page={{this.page}} @size={{this.size}} @class="container-flex--contain" as |t|>
-  <t.menu as |menu|>
+<AuDataTable @content={{this.model}} @isLoading={{this.isLoadingModel}} @noDataMessage="Geen zittingen" @sort={{this.sort}} @page={{this.page}} @size={{this.size}} @class="container-flex--contain" as |table|>
```
Windvis, line 1: Pro tip, if you start seeing weird rendering errors after adding translations, check that there is no local block variable called `t`. Took me half an hour of debugging before I figured it out :smile:.

app/templates/inbox/meetings/new.hbs
```diff
@@ -0,0 +1,70 @@
+{{!-- TODO: Uncomment this once the app uses `{{page-title}}` on all relevant pages --}}
+{{!-- {{page-title (t "inbox.meetings.new.title")}} --}}
```
Windvis, line 2: Using this overrides the title that's set when the app boots. I think page titles are useful though. Could be a nice improvement to add this to all relevant pages in the future?

app/templates/inbox/meetings/new.hbs
```diff
@@ -0,0 +1,70 @@
...
+>
+  <Modal.Body>
+    <form class="au-o-flow" id="create-meeting-form" {{on "submit" (perform this.saveMeetingTask)}}>
+      <div>
+        {{#let (unique-id) this.meeting.errors.bestuursorgaan as |id errors|}}
+          <AuLabel
+            @error={{if errors true false}}
+            for={{id}}
+          >
+            {{t "inbox.meetings.new.meeting.administrativeBody"}} <AuPill>Verplicht</AuPill>
```
Windvis, line 18: Forgot to translate the "Verplicht" part. But I was thinking that a small template only component would make more sense? I'm guessing there will be other places where it's needed.

app/templates/inbox/meetings/new.hbs
```diff
@@ -0,0 +1,70 @@
...
+            <AuHelpText @error={{true}}>
+              {{t error.message}}
+            </AuHelpText>
+          {{/each}}
+        {{/let}}
+      </div>
+      <div>
+        {{#let (unique-id) as |id|}}
+          <AuLabel for={{id}}>{{t "inbox.meetings.new.meeting.location"}}</AuLabel>
+          <AuInput @value={{this.meeting.opLocatie}} @width="block" id={{id}} type="text" />
```
Windvis, line 36: I took the liberty of moving this up (so the date inputs are together) and making it full width. I think it looks better :smile:. Let me know if that isn't ok.

app/templates/inbox/meetings/new.hbs
```diff
@@ -0,0 +1,70 @@
+{{!-- TODO: Uncomment this once the app uses `{{page-title}}` on all relevant pages --}}
+{{!-- {{page-title (t "inbox.meetings.new.title")}} --}}
+
+<AuModal
```
Windvis, line 4: I was planning on extracting a modal component that could be used both here and on the edit page. But while implementing that I noticed that the conditional logic, component arguments, and hacking around required to make it work were not worth it. The template isn't very DRY, but I think simplicity and flexibility this gives is worth it. Now the creation modal can be changed independently. (I don't think all the date fields are needed when creating a meeting for example).

app/controllers/inbox/meetings/new.js
```diff
@@ -0,0 +1,42 @@
...
+  @action
+  updateAdministrativeBody(administrativeBody) {
+    this.meeting.bestuursorgaan = administrativeBody;
+    this.meeting.errors.remove('bestuursorgaan');
+  }
+
+  @action
+  cancelMeetingCreation() {
+    // Prevent unwanted calls due to focus trap...
+    if (this.meeting.isNew && !this.meeting.isDeleted) {
```
Windvis, line 37: the ember-focus-trap double execution issue bit me again. I PR'd a fix for it a while ago, but that only fixes the issue if you also change the `@modalOpen` argument of the AuModal component. Since it's statically set to `true` in this route, the issue still occurs. I think this is a deeper issue in the focus-trap modifier itself but I haven't figured it out the fix yet.

app/routes/inbox/meetings/new.js
```diff
@@ -0,0 +1,26 @@
...
+
+  model() {
+    let now = new Date();
+    let meeting = this.store.createRecord('zitting', {
+      geplandeStart: now,
+      gestartOpTijdstip: now,
+    });
+
+    return {
+      meeting,
```
Windvis, line 23: Not conventional but I wanted to get the conversation started:

I was thinking that _always_ returning an object (or a promise resolving to an object) from the model hook is a good default even if you only return 1 thing. I think it has 2 benefits compared to the alternative:
1. It sidesteps the issue that `@model` isn't very descriptive. `@model.meeting` is way more readable, at the cost of extra typing. That also means you don't _have_ to create a controller to alias things (or use the `{{let}}` helper to do it).
2. Once you do need extra model props in the route, you can simply add them, without having to start mass renaming all your `model` usage to include properties.

Now there are other solutions as well, but I think this kind of makes sense as a guideline since it doesn't stray away from the Ember way of doing things?

Opinions?

app/components/administrative-body-select.hbs
```diff
@@ -0,0 +1,18 @@
+<div class={{if @error "ember-power-select--error"}}>
```
Windvis, line 1: Not sure if that's the proper way to add an error state to PowerSelect. I couldn't find it in the docs and looked in the styles instead. It looks a bit weird since it's a BEM modifier though but the `ember-power-select` class doesn't seem to include styles. So I would only add it to follow the BEM rules. Renaming the class might make more sense though.

app/components/zitting/manage-zittingsdata.hbs
```diff
@@ -8,24 +30,19 @@
...
-            @placeholder={{t "manageZittingsData.selectPlaceholder"}}
-            @selected={{this.bestuursorgaan}}
-            @options={{this.bestuursorgaanOptions}}
-            @onChange={{this.changeSelect}}
-            as |bestuursorgaan|>
-              {{bestuursorgaan.isTijdsspecialisatieVan.naam}}
-              (periode: {{moment-format bestuursorgaan.bindingStart "MM/DD/YYYY"}} -
-              {{#if bestuursorgaan.bindingEinde}}{{moment-format bestuursorgaan.bindingEinde "MM/DD/YYYY"}}{{else}}nvt.{{/if}})
-          </PowerSelect>
+          {{this.bestuursorgaan.isTijdsspecialisatieVan.naam}}
```
Windvis, line 33: This is now simply displayed as plain text, which might not be the best idea. Should I show it as a disabled input field (or select) instead?

app/controllers/inbox/meetings/new.js
```diff
@@ -0,0 +1,42 @@
...
+      this.meeting.destroyRecord();
+      this.router.replaceWith('inbox.meetings');
```
Windvis, line 39: 1 thing I noticed is that the loading template is shown very briefly when returning back to the `inbox.meetings` route. It looks a bit glitchy / weird, but it's related to the `refreshModel` option of the query params so I don't think I can fix it myself. 

AFAIK the model hooks shouldn't be rerun when transitioning between sibling routes (.new to .index), and the query params don't change either, so I'm not sure if that's a router bug or not.

Windvis: Someone already created an issue for it: https://github.com/emberjs/ember.js/issues/19497

app/components/meeting-form.hbs
```diff
@@ -92,7 +92,10 @@
   <div class="au-c-meeting-chrome">
     <div class="au-c-meeting-chrome__paper">
       {{!-- Meeting title --}}
-      <AuHeading>{{t "meetingForm.meetingHeadingPartOne"}} <span class="au-c-meeting-chrome__highlight">{{@zitting.bestuursorgaan.isTijdsspecialisatieVan.naam}}</span>, {{t "meetingForm.meetingHeadingPartTwo"}} <span class="au-c-meeting-chrome__highlight">{{moment-format  @zitting.geplandeStart "DD/MM/YYYY HH:mm"}}</span></AuHeading>
+      <AuHeading>
+        {{t "meetingForm.meetingHeadingPartOne"}} <span class="au-c-meeting-chrome__highlight">{{@zitting.bestuursorgaan.isTijdsspecialisatieVan.naam}},</span>
```
Windvis, line 96: I moved the `,` to be inside the highlight because if the name is long enough to wrap multiple lines, the comma floats on the line below, which looks funky:

image

app/components/zitting/manage-zittingsdata.hbs
```diff
@@ -1,5 +1,27 @@
-<div {{did-insert this.fetchBestuursorgaan}}>
-  {{#if this.showModal}}
+<div class="au-c-meeting-chrome-card au-u-margin-bottom-large">
```
Windvis, line 1: This content is now always shown, even when the modal is open. That prevents the background from jumping around when opening or closing the modal.

app/templates/inbox/meetings/new.hbs
Windvis: It should be possible to uncomment that once #118 is merged.

app/controllers/inbox/meetings/new.js
Windvis: https://github.com/appuniversum/ember-appuniversum/pull/111 should fix this issue.

app/templates/inbox/meetings/new.hbs
nvdk: I've merged #118 :)

app/components/administrative-body-select.hbs
nvdk: tbh I have no idea, perhaps @Dietr  has some pointers

app/components/administrative-body-select.js
```diff
@@ -0,0 +1,53 @@
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
+
+const VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS = [
```
nvdk, line 6: I see you've hardcoded this now, not sure I'm a fan of that. I have to admit this is not data that changes often.

nvdk: nm, I see it was already hardcoded. oh well :)

app/routes/inbox/meetings/new.js
nvdk: not sure if I'm a fan, mostly because it makes it slightly harder to provide to model to a route with linkto (though we don't do that often).

app/templates/inbox/meetings/new.hbs
```diff
@@ -0,0 +1,70 @@
...
+          <AuLabel for={{id}}>{{t "inbox.meetings.new.meeting.location"}}</AuLabel>
+          <AuInput @value={{this.meeting.opLocatie}} @width="block" id={{id}} type="text" />
+        {{/let}}
+      </div>
+      <div>
+        <AuLabel>{{t "inbox.meetings.new.meeting.plannedStart"}}</AuLabel>
+        <DateTimePicker @onChange={{fn (mut this.meeting.geplandeStart)}} @value={{this.meeting.geplandeStart}}/>
+      </div>
+      <div>
+        <AuLabel>{{t "inbox.meetings.new.meeting.start"}}</AuLabel>
```
nvdk, line 44: since we're improving the flow anyway and given that this is now separated I'd actually drop the start and end datetime pickers, as they only make sense after the meeting has happened

app/components/zitting/manage-zittingsdata.hbs
nvdk: I think we use both patterns at the moment, so for me it's ok.

app/routes/inbox/meetings/new.js
Windvis: You mean providing the actual model instead of the record id? Yea, I didn't think of that, but I don't think that's a drawback :smile:.

Anyway, I figure you prefer to return the record directly instead?

app/components/administrative-body-select.js
Windvis: Yea I just moved it from the other component and defined it as a const instead. If you give me pointers on how to do it without hardcoding I can take a look at it if you want :smiley:.

app/templates/inbox/meetings/new.hbs
Windvis: Yea, I agree, that's one of the reasons I didn't create a single component that covers both cases. I had a gut feeling that they "looked the same, but aren't the same". I'll remove those fields.

app/routes/inbox/meetings/new.js
nvdk: I have a slight preference for having the record directly with an alias indeed.

app/controllers/inbox/meetings/new.js
Windvis: Removed the logic since it's no longer needed now.

app/routes/inbox/meetings/new.js
Windvis: Updated.

app/templates/inbox/meetings/new.hbs
Windvis: Created a simple `<RequiredField>` template only component for this. Let me know if it needs to be named differently :smile:.

PR #118, opened 2021-05-28
Set the application title using `{{page-title}}`

Windvis:
This sets the application title using the `{{page-title}}` helper. This allows other pages to also use the helper without overriding the base title.


app/templates/application.hbs
```diff
@@ -1,5 +1,12 @@
 <EmberLoadRemover />
 
+{{#let (config "environmentName") as |environmentName|}}
```
Windvis, line 3: Such a useful helper :smile: :tada:

PR #120, opened 2021-06-01
Feature/editor modal

abeforgit:
fixes https://binnenland.atlassian.net/browse/GN-2637

I tried but couldn't figure out how to show a blocking modal that wasnt the window.confirm builtin.
There's also quite a bit of duplication in the new routes, but at least there is also reduced duplication because of a refactor of rdfa-editor-container, and this already took way too long.


abeforgit:
I forgot to add read-only checks for the routes, ill add them in a bit

tests/unit/controllers/meetings/edit/intro-test.js
```diff
@@ -0,0 +1,12 @@
+import { module, test } from 'qunit';
```
nvdk, line 1: please remove empty tests

app/controllers/meetings/edit/intro.js
```diff
@@ -0,0 +1,63 @@
...
+  removeExitHandler() {
+    this.router.off('routeWillChange', this, this.confirm);
+  }
+
+  confirm(transition) {
+    if (transition.isAborted) {
+      return;
+    }
+    if (this.dirty &&
+      !window.confirm(
```
nvdk, line 56: for this case I don't particularly mind the window.confirm actually

app/templates/meetings/edit/outro.hbs
```diff
@@ -0,0 +1,29 @@
...
+  <Modal.Body>
+    <ZittingTextDocumentContainer
+            @onEditorInit={{this.initEditor}}
+            @saving={{this.saveTextTask.isRunning}}
+            @type="ext:outro"
+            @zitting={{@model}}/>
+  </Modal.Body>
+  <Modal.Footer>
+    <AuButtonGroup>
+      <AuButton @skin="secondary"
```
nvdk, line 15: personally, I'd just disable the buttons when saveTextTask is running and include a spinner next to the buttons.

nvdk (commented):
seems to be working ok, so at least all that work payed off! based on a quick review looks good, could use a bit of designer love though :)

app/routes/meetings/edit/treatment.js
```diff
@@ -0,0 +1,43 @@
...
+
+export default class MeetingsEditTreatmentRoute extends Route {
+  @service intl;
+
+  async model(params) {
+    const treatment = await this.store.findRecord(
+      'behandeling-van-agendapunt',
+      params.treatment_id
+    );
+    console.log('ROUTER PARAMS', params);
```
nvdk, line 13: debug statement can go

app/components/behandeling-van-agendapunt.hbs
```diff
@@ -1,102 +1,87 @@
...
-  {{/if}}
+
+<ReadonlyTextBox @title={{t "behandelingVanAgendapunten.contentTitle"}}>
+  {{@behandeling.documentContainer.currentVersion.htmlSafeContent}}
+</ReadonlyTextBox>
+{{#if this.editable}}
   <AuButtonGroup>
-    <AuButton {{on "click" this.save}} @disabled={{if this.saveEditorDocument.isRunning true false}}>
-      {{t "behandelingVanAgendapunten.saveButton"}}
+    <AuButton {{on "click" this.goToEdit}}>
```
nvdk, line 83: this could just be a LinkTo perhaps ?

app/controllers/meetings/edit/intro.js
```diff
@@ -0,0 +1,63 @@
...
+
+  @task
+  *saveTextTask() {
+    const zitting = this.model;
+    zitting.intro = this.editor.htmlContent;
+    yield zitting.save();
+  }
+
+  addExitHandler() {
+    this.router.on('routeWillChange', this, this.confirm);
```
Windvis, line 44: I've done something similar in the past and instead of hooking this up in all the controllers I created a component to do it for me. Back then I kept it generic so I went for `<WillTransition>` but I think a more opinionated component might be better since this is one of  the only use-cases I can think of (and that would remove some extra setup code as well).

An API similar to this might be nice?

```hbs
<ConfirmRouteLeave @enabled={{this.dirty}} @message={{t "meetings.edit.intro.confirmQuitWithoutSaving"}}  />
```

That way the controllers would only need to store the dirty state and all the other setup code can be removed. I _think_ that should work unless I'm missing something.

app/components/meeting-form.hbs
```diff
@@ -161,8 +161,12 @@
                 {{t 'manageFreeText.before'}}
             </span>
           </AuHeading>
-          <ZittingTextDocumentContainer @zitting={{@zitting}} @readOnly={{this.readOnly}} @type="ext:intro"/>
-
+        <ReadonlyTextBox @title={{t "behandelingVanAgendapunten.contentTitle"}}>
+          {{@zitting.intro}}
```
Windvis, line 165: I think these need a `html-safe` wrapper?

app/controllers/meetings/edit/intro.js
Windvis: Yea and "modal on modal" isn't supported by Appuniversum either I think.

app/components/behandeling-van-agendapunt.js
```diff
@@ -6,11 +6,13 @@ import { action } from "@ember/object";
 import { inject as service } from "@ember/service";
 
 export default class BehandelingVanAgendapuntComponent extends Component {
+  @service store;
+  @service currentSession;
```
Windvis, line 10: This service doesn't seem to be used (anymore)?

app/components/rdfa-editor-container.js
```diff
@@ -1,29 +1,49 @@
...
-      return {
-        prefix: "",
-        typeof: "",
-        vocab: ""
-      };
+    if (this.args.editorDocument) {
+      try {
+        return JSON.parse(this.args.editorDocument.context);
+      } catch (e) {
+        return {
```
Windvis, line 35: I _think_ this can be dropped since it's already returned after the try catch?

app/controllers/meetings/edit/treatment.js
```diff
@@ -0,0 +1,136 @@
...
+  @tracked initialContent;
+  @tracked confirmLeaveOpen;
+
+  get dirty() {
+    return this.document.content !== this.editor.htmlContent;
+  }
+
+  @action
+  closeModal() {
+    this.router.transitionTo('meetings.edit', this.model.meetingId);
```
Windvis, line 25: Passing the meetingId here shouldn't be needed since the router has that data already (you just go up the stack).

app/components/rdfa-editor-container.js
```diff
@@ -35,12 +55,23 @@ export default class RdfaEditorContainerComponent extends Component {
    */
   @action
   setPrefix(element) {
-    element.setAttribute('prefix', this.prefixToAttrString(this.documentContext.prefix));
+    element.setAttribute(
+      'prefix',
```
Windvis, line 59: I know this code was already there, but I don't understand why `prefix` must be set through a modifier?

app/controllers/meetings/edit/treatment.js
```diff
@@ -0,0 +1,136 @@
...
+
+  @action
+  async saveAndQuit() {
+    await this.saveDocumentTask.perform();
+    this.closeModal();
+  }
+  @action
+  async cancel() {
+    const container = await this.documentContainer;
+    container.rollbackAttributes();
```
Windvis, line 44: I don't think the rollback is needed here since the confirm behavior will handle that?

app/routes/meetings/edit/treatment.js
```diff
@@ -0,0 +1,43 @@
...
+      treatment,
+      meetingId: this.paramsFor('meetings.edit').id,
+    };
+  }
+
+  @action
+  willTransition(transition) {
+    //
+
+    // the ember official docs contain an example where the controller is accessed in the route
```
Windvis, line 24: I would go with the same setup as the other routes?

app/templates/meetings/edit/treatment.hbs
```diff
@@ -0,0 +1,35 @@
...
+        {{t "behandelingVanAgendapunten.saveButton"}}
+      </AuButton>
+      <AuButton @loading={{this.saveDocumentTask.isRunning}} {{on "click" this.saveAndQuit}}>
+        {{t "behandelingVanAgendapunten.saveAndQuit"}}
+      </AuButton>
+    </AuButtonGroup>
+    <AuButtonGroup>
+       <AuButton @skin="secondary"
+                @loading={{this.saveDocumentTask.isRunning}}
+        {{on "click" this.cancel}}>
```
Windvis, line 30: Any specific reason why you split `cancel` and `closeModal` up into separate actions?

translations/en-us.yaml
```diff
@@ -36,6 +36,10 @@ publish:
 behandelingVanAgendapunten:
   documentLink: "document link"
   saveButton: Save
+  saveAndQuit: Save and Quit
+  cancel: Cancel
+  leaveWithoutSaving: Leave wihout saving
```
Windvis, line 41: `without` typo

Windvis:
I made a lots of comments but most of them are only nitpicky things, I think you did a great job! Nice refactoring work as well!

app/components/rdfa-editor-container.js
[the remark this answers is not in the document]
nvdk: see https://github.com/emberjs/ember.js/issues/19369

app/components/behandeling-van-agendapunt.hbs
abeforgit: at the time the version of AU we have didnt support models with multiple params in AuLink
I see that theres been a new release so Ill try it out

Windvis: FYI I think you only need to pass the treatment id since the router already has the meeting id.

Windvis: Hmm since it's a component that in theory could be put anywhere passing everything might make sense actually.

app/controllers/meetings/edit/intro.js
abeforgit: i almost did this as well, but I found it a weird usecase for a component. So I tried with a helper, but that didn't have a destructor, so I couldn't unregister the listener. And then I gave up because I was going mad

app/components/behandeling-van-agendapunt.hbs
abeforgit: at some point ember complained so I just assumed I had to put everything everywhere
but it does seem to work with only 1 argument

tests/unit/controllers/meetings/edit/intro-test.js
abeforgit: should I just remove all tests then? they're all empty
or only the ones I added?

app/templates/meetings/edit/outro.hbs
abeforgit: I don't even think we need the spinner, we already have the scanner animation

tests/unit/controllers/meetings/edit/intro-test.js
nvdk: feel free to remove all empty tests :+1:

app/controllers/meetings/edit/intro.js
Windvis: Why do you think it's a weird use-case for it? Hmm, a class base helper should have a `willDestroy` hook, but I can't find it in the API docs. Not sure if that's a bug in the docs or not.

app/templates/meetings/edit/treatment.hbs
abeforgit: I believe the technical term for it is "being an idiot"

app/controllers/meetings/edit/intro.js
abeforgit: Well it doesn't render anything
so the template is essentially a useless file
so id want a component without a template
which to me is basically a class based helper
but yeah I didnt see the hook in the docs, so I gave up there

app/components/behandeling-van-agendapunt.hbs
abeforgit: Visually the AuLink looks very different than the button though, and with the changes dieter made I think the button looks more uniform (it's what we use for the other, non-fullpage modals)

app/controllers/meetings/edit/intro.js
abeforgit: also incidentally, if it does have a willDestroy hook, how do you know this? I mean I've accepted you are an ember wizard, but where did you get your powers? :stuck_out_tongue:

app/components/behandeling-van-agendapunt.hbs
Windvis: I would just use a `LinkTo` and add the needed classes. I think Appuniversum should simply provide a component that makes an `<a>` tag look like a button since it's such a common use-case. I already created an issue for that: https://github.com/appuniversum/ember-appuniversum/issues/110 :smile: 

Using button for navigation just because they look like buttons seems weird to me :smile:.

app/controllers/meetings/edit/intro.js
Windvis: I've used it myself :smile: (which doesn't mean it's public API, I know), but I've seen official snippets where they use it like that. Which makes me assume it's simply a documentation bug. I'll try to find an official reference later.

Windvis: Well the same can be said for helpers that don't return a value :smile:. I think the framework just doesn't cover that use-case yet. "effects" _might_ solve it in the future but that's not implemented yet. Personally I don't have issues with components that don't really use the template. They can also simply serve as reusable logic. But maybe I'm biased because I've done it like that already.

Anyway I still think it's better than repeating this logic in all the controllers that need it :smile:. (I think the "new meeting" page might also want to use this in the future for example).

abeforgit: yeah I think ill go for the component approach
rather a bit of a weird component than an undocumented helper hook I think

app/components/behandeling-van-agendapunt.hbs
abeforgit: well yeah
but this story also need to get done at some point :stuck_out_tongue:

app/controllers/meetings/edit/intro.js
abeforgit: alright done
a lot better for sure

app/routes/meetings/edit/treatment.js
```diff
@@ -0,0 +1,43 @@
...
+      transition.abort();
+    } else {
+      controller.documentContainer.rollbackAttributes;
+      return true;
+    }
+  }
+
+  setupController(controller, model) {
+    super.setupController(controller, model);
+    controller.editor = null;
```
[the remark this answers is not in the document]
abeforgit: good thinking, changed it

app/components/behandeling-van-agendapunt.hbs
```diff
@@ -1,102 +1,85 @@
...
-  </AuButtonGroup>
-</div>
-{{/if}}
\ No newline at end of file
+
+<ReadonlyTextBox @title={{t "behandelingVanAgendapunten.contentTitle"}}>
+  {{@behandeling.documentContainer.currentVersion.htmlSafeContent}}
+</ReadonlyTextBox>
+{{#if this.editable}}
+  <AuButton @width="block" @skin="secondary" @icon="pencil" @iconAlignment="left" {{on "click" this.goToEdit}}>
```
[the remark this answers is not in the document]
abeforgit: see above, waiting for Appuniversum support

tests/unit/controllers/meetings/edit/intro-test.js
abeforgit: https://github.com/lblod/frontend-gelinkt-notuleren/pull/122

abeforgit:
> I forgot to add read-only checks for the routes, ill add them in a bit

these are added now

app/components/behandeling-van-agendapunt.hbs
nvdk: I've created a story on the backlog, so we can take this up at a later time

app/controllers/meetings/edit/intro.js
Windvis: https://github.com/emberjs/rfcs/blob/master/text/0626-invoke-helper.md I found this RFC which includes a class based helper with a `willDestroy` hook, btw. There are probably more references, but that's the first "official" one I encountered :smile:. I also created an issue to update the API docs.

PR #121, opened 2021-06-02
Feature/sort voting

lagartoverde:
Added new position property to votings and sort on that on the frontend

### DIFF
--- a/app/components/treatment/voting/modal.js
+++ b/app/components/treatment/voting/modal.js
@@ -33,18 +33,19 @@ export default class TreatmentVotingModalComponent extends Component {
   @restartableTask
   /** @type {import("ember-concurrency").Task} */
   fetchStemmingen = function* () {
-    this.stemmingen = yield this.args.behandeling.stemmingen;
+    this.stemmingen = (yield this.args.behandeling.stemmingen).sortBy('index');
   };
 
   @task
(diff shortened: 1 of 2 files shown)


app/components/treatment/voting/modal.js
```diff
@@ -33,18 +33,19 @@ export default class TreatmentVotingModalComponent extends Component {
...
   /** @type {import("ember-concurrency").Task} */
   saveStemming = function* () {
     const isNew = this.editStemming.stemming.isNew;
-    yield this.editStemming.saveTask.perform();
     if (isNew) {
+      this.editStemming.stemming.position = this.args.behandeling.stemmingen.length;
       this.args.behandeling.stemmingen.pushObject(this.editStemming.stemming);
       this.args.behandeling.save();
     }
+    yield this.editStemming.saveTask.perform();
```
nvdk, line 48: I think this needs to be before the save of the behandeling, not a 100% sure

lagartoverde: I tested and it worked fine, but I can leave it before, I just changed it in order to be a little bit more concise because I need to put 2 if isNew statements if not, I will change it just to be sure

abeforgit:
Looks good!

abeforgit (approved):

PR #122, opened 2021-06-02
Remove generated tests

abeforgit:
Nobody's running them, nobody's updating them, so let's get rid of them


Windvis:
https://imgflip.com/i/5bug29

PR #128, opened 2021-06-03
Feature/flexible print

abeforgit:
Implements https://binnenland.atlassian.net/browse/GN-2649

Also refactors the print page into a reusable component

This allows printing unsigned extracts (which don't exist in the backend) to be printed. This is implemented by passing the entire model in the transition rather than just the id (which it of course doesn't have). It works quite well, but it breaks the refresh button and has a somewhat ugly url ending in `null`. To be decided if this is an issue.

This also extracts the refactored route template into a component. This was originally to accomodate rendering the printview in a modal to allow for an unsaved document to be printed. I ended up choosing the abovementioned approach instead, cause I wasn't sure what rendering inside a modal would do for the browser's builtin printing feature, and I thought it cleaner to keep it on one page.

If it turns out the current implementation is sufficient, I can easily unwrap the component back into the route template. There may also be some approach using query templates that could work, but for that I'd need to do more research, and this already worked so I didn't look into it.


nvdk:
Ah yes, quite the bit of complexity we missed when estimating once again!
Given that it's now a component, I think it would make sense to split it up into two routes? this would also allow you to move the fetching of the preview to the route itself.

abeforgit:
we already have the preview at that point though
and its not so simple because a meeting has multiple extracts, so the new route would also need some kind of identifier

abeforgit:
ah but I guess we could use the id for the behandeling then?
since an extract is basically a behandeling right?

nvdk:
Yeah so not super clean, but you could pass the preview with the id set to behandeling id (or whatever we need to generate the preview, but I presume the behandelling). That means you could have the model hook fetch the preview, but still allows you to pass in the preview if you already have it.

abeforgit:
Blood, sweat and mostly tears as usual.

I think it's a decent implementation now? I honestly can't tell anymore.
Basically the service acts as a domain-specific store without implicit reloading. So we get a merged map of "real (aka persisted)" versionedBehandeling objects and "previews", keyed by the id of the treatment they represent. In the print route we then fetch the preview by the id of the treatment, and we don't care whether this is a persisted one or a preview (well we care in the sense that we can handle both in the template). If the lookup fails, we do a reload and try again (this handles refreshing the page for example). Actually now that I write this down, we could do that more elegantly by only fetching the specific preview we care about, and creating it if it doesn't exist. 

Issues can probably appear since this is basically doing ad-hoc caching, depending on the developer to reload the cache. I don't know if this is a good pattern or not. It seems to be in the spirit of what ember-storefront proposes, but I might be wrong about that.

nvdk:
@abeforgit I see you reopened this, how does it relate to #134 ?

Windvis:
@nvdk The other PR uses this one as the target so I think it's split up to make everything easier to review?

PR #129, opened 2021-06-04
Feature/link ap intermission

Asergey91:
solves GN-2658


app/components/manage-intermissions/edit.js
```diff
@@ -79,4 +83,49 @@ export default class manageIntermissionsEditComponent extends Component {
     yield intermission.destroyRecord();
     this.args.onClose();
   }
+
+  //position stuff
+  @tracked locationOptions = [
+    { code: 'during', name: 'Tijdens agendapunt' },
```
lagartoverde, line 89: I think this should use the i18n plugin to provide strings both in Dutch and English

app/components/manage-intermissions/index.hbs
```diff
@@ -33,6 +36,20 @@
...
+                {{#if (eq intermission.position "during")}}
+                  {{t "manageIntermissions.duringAp"}}
+                {{/if}}
+                {{#if (eq intermission.position "before")}}
+                  {{t "manageIntermissions.beforeAp"}}
+                {{/if}}
+                {{#if (eq intermission.position "after")}}
+                  {{t "manageIntermissions.afterAp"}}
+                {{/if}}
+                <strong>{{intermission.onderwerp.titel}}</strong>
```
lagartoverde, line 50: How hard would it be to make this a link to the specified behandeling? I saw as bold and instinctively I clicked it thinking it will bring be to the behandeling. Maybe we should discuss the feature but I think is a nice addition

lagartoverde (approved):
Small remark about the translations and a comment about a possible feature, but everything looks good, good job 👍 .

app/components/manage-intermissions/edit.js
nvdk: indeed, there's a `t` helper you can import and use

app/components/manage-intermissions/index.hbs
abeforgit: not trivial I think, we don't really do anything with url fragments yet
but I think it's a really good idea, also in the gebruikersessie we saw people clicking on various things expecting they were links

abeforgit (commented):
Small nitpicks, you know me
But works well and looks good otherwise!

app/components/manage-intermissions/index.hbs
lagartoverde: But the side menu has the links, for example http://localhost:4200/meetings/60B9D8A10AAB420009000001/edit#behandeling-60B9D8CC0AAB420009000004 goes to the first behandeling, what do you mean with url fragments? the hashes?

lagartoverde: I think the clicking on this also has to do with the style of the bold, something to check with dieter I think

nvdk: it works as long as you remain on the same route, non trivial otherwise

app/components/manage-intermissions/edit.js
```diff
@@ -79,4 +83,49 @@ export default class manageIntermissionsEditComponent extends Component {
...
+      this.args.intermissionToEdit.position = null;
+      this.selectedAp = null;
+    }
+  }
+
+  get selectedAp() {
+    return this.args.intermissionToEdit.onderwerp;
+  }
+  set selectedAp(value) {
+    this.args.intermissionToEdit.position = this.selectedLocation.code;
```
nvdk, line 119: as mentioned in the backend PR (https://github.com/lblod/app-gelinkt-notuleren/pull/29 ) it might make sense to capture these in a separate model "DocumentPosition" or something similar.

app/components/manage-intermissions/edit.js
```diff
@@ -79,4 +84,78 @@ export default class manageIntermissionsEditComponent extends Component {
...
+  get positionOptions() {
+    return [
+      { code: 'before', name: this.intl.t('manageIntermissions.beforeAp'), conceptUuid: "9c9be842-236f-4738-b642-f4064c86db51" },
+      { code: 'during', name: this.intl.t('manageIntermissions.duringAp'), conceptUuid: "4790eec5-acd2-4c1d-8e91-90bb2998f87c" },
+      { code: 'after', name: this.intl.t('manageIntermissions.afterAp'), conceptUuid: "267a09cc-5380-492d-93ad-697b9e99f032" }
+    ];
+  }
+  @action
+  updatedIntermission(){
+    this.args.intermissionToEdit;
```
nvdk, line 98: I don't think this statement does anything and it can probably go?

app/components/manage-intermissions/edit.hbs
```diff
@@ -1,3 +1,5 @@
+
+<div {{did-update this.updatedIntermission @intermissionToEdit}}>
```
nvdk, line 2: no need to pass in `@intermissionToEdit` here I think

app/components/manage-intermissions/edit.js
```diff
@@ -79,4 +84,78 @@ export default class manageIntermissionsEditComponent extends Component {
     yield intermission.destroyRecord();
     this.args.onClose();
   }
+
+  //position stuff
+  get positionOptions() {
+    return [
+      { code: 'before', name: this.intl.t('manageIntermissions.beforeAp'), conceptUuid: "9c9be842-236f-4738-b642-f4064c86db51" },
+      { code: 'during', name: this.intl.t('manageIntermissions.duringAp'), conceptUuid: "4790eec5-acd2-4c1d-8e91-90bb2998f87c" },
+      { code: 'after', name: this.intl.t('manageIntermissions.afterAp'), conceptUuid: "267a09cc-5380-492d-93ad-697b9e99f032" }
```
nvdk, line 93: I'd recommend defining the uuids as constants so they can have a meaningful name

Asergey91: I think actually just passing the concept scheme here instead is the better solution

app/components/manage-intermissions/edit.hbs
Asergey91: tried without it, doesnt work

PR #134, opened 2021-06-08
Feature/publish service

abeforgit:
Blood, sweat and mostly tears as usual.

I think it's a decent implementation now? I honestly can't tell anymore.
Basically the service acts as a domain-specific store without implicit reloading. So we get a merged map of "real (aka persisted)" versionedBehandeling objects and "previews", keyed by the id of the treatment they represent. In the print route we then fetch the extract by the id of the treatment, and we don't care whether this is a persisted one or a preview (well we care in the sense that we can handle both in the template). If the lookup fails, we do a reload and try again (this handles refreshing the page for example). Actually now that I write this down, we could do that more elegantly by only fetching the specific preview we care about, and creating it if it doesn't exist. 

Issues can probably appear since this is basically doing ad-hoc caching, depending on the developer to reload the cache. I don't know if this is a good pattern or not. It seems to be in the spirit of what ember-storefront proposes, but I might be wrong about that.

### DIFF
--- a/app/controllers/meetings/publish/uittreksels.js
+++ b/app/controllers/meetings/publish/uittreksels.js
@@ -1,106 +1,72 @@
 import Controller from '@ember/controller';
-import {task} from "ember-concurrency";
-import {tracked} from "@glimmer/tracking";
+import { task } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
 import { fetch } from 'fetch';
 import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
 
 export default class MeetingsPublishUittrekselsController extends Controller {
-  @tracked uittreksels;
+  @service publish;
   @tracked documentToPrint;
   @tracked showPrintModal;
 
   @tracked errors;
 
-
   initialize() {
-    this.uittreksels = [];
     this.documentToPrint = null;
     this.showPrintModal = false;
-    this.initializeUittreksels.perform();
+    this.loadExtractsTask.perform();
   }
 
   get meeting() {
     return this.model;
   }
 
-  @task
-  * initializeUittreksels() {
-    const uittreksels = [];
-    const previews = yield this.fetchExtractPreviews.perform();
-    for(const uittreksel of previews) {
-      const existingUittreksels = yield this.store.query('versioned-behandeling',{
-        'filter[behandeling][:id:]': uittreksel.data.attributes.uuid,
-        include: 'signed-resources,published-resource,behandeling'
-      });
-      if(existingUittreksels.length) {
-        uittreksels.push({document: existingUittreksels.firstObject});
-      } else {
-        const behandeling = yield this.store.findRecord('behandeling-van-agendapunt', uittreksel.data.attributes.uuid);
-        const rslt = yield this.store.createRecord("versioned-behandeling", {
-          zitting: this.meeting,
-          content: uittreksel.data.attributes.content,
-          behandeling,
-        });
-        uittreksels.push({document: rslt, errors: uittreksel.data.attributes.errors});
-      }
-    }
-    this.uittreksels = uittreksels;
+  get uittreksels() {
+    return this.publish.treatmentExtracts;
+  }
+
+  get isLoading() {
+    return this.loadExtractsTask.isRunning;
   }
 
   @task
-  * reloadUittreksels() {
-    const uittreksels = [];
-    const previews = yield this.fetchExtractPreviews.perform();
-    for(const uittreksel of previews) {
-      const existingUittreksels = yield this.store.query('versioned-behandeling',{
-        'filter[behandeling][:id:]': uittreksel.data.attributes.uuid,
-        include: 'signed-resources,published-resource'
-      });
-      if(existingUittreksels.length) {
-        uittreksels.push(existingUittreksels.firstObject);
-      } else {
-        const behandeling = yield this.store.findRecord('behandeling-van-agendapunt', uittreksel.data.attributes.uuid);
-        const rslt = yield this.store.createRecord("versioned-behandeling", {
-          zitting: this.model,
-          content: uittreksel.data.attributes.content,
-          behandeling,
-        });
-        uittreksels.push(rslt);
-      }
-    }
-    this.uittreksels = uittreksels;
+  *loadExtractsTask() {
+    yield this.publish.loadExtractsTask.perform(this.meeting.id);
   }
 
+  get createSignedResourceTask() {
+    return this._createSignedResourceTask.unlinked();
+  }
 
-  @task
-  *fetchExtractPreviews() {
-    const response = yield fetch(`/prepublish/behandelingen/${this.meeting.id}`);
-    const json = yield response.json();
-    return json;
+  get createPublishedResourceTask() {
+    return this._createPublishedResourceTask.unlinked();
   }
 
   @task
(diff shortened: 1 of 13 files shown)


app/services/publish.js
```diff
@@ -0,0 +1,106 @@
...
+      treatmentMap.set(treatment.id, treatment)
+    );
+
+    // map all existing extracts to their behandeling id
+    const extractMap = new Map();
+    versionedTreatments.forEach((versionedTreatment) => {
+      extractMap.set(
+        versionedTreatment.behandeling.get('id'),
+        new Extract(
+          versionedTreatment.behandeling.get('id'),
```
abeforgit, line 73: I have a feeling this code is not sound. It works because behandeling is included, but I don't know if it is guaranteed to work, since technically this is still a relationship and should thus be treated as being async.

I think the proper way to solve this would be to yield the behandeling, and rely on the include to avoid the unnecessary network request. However, since nothing in ember data is iterable (that would just be sensible, and of course we don't ever want anything sensible) this is really annoying to do (you can't yield from a forEach).

@nvdk @Windvis any suggestions?

Windvis: Always resolving the async hasMany relationship before accessing its data does seem to be the best practice. I've only recently learned of this myself since there is an [open RFC](https://github.com/runspired/rfcs/blob/ember-data/deprecate-methods-on-promise-many-array/text/0745-ember-data-deprecate-methods-on-promise-many-array.md) to deprecate the direct data access. The RFC only mentions hasMany relationships though.. not sure if that's different from belongsTo (and I don't really know how to find that out either besides asking in the Discord). I'm still far away from being an Ember Data expert, sorry :smile:.

> since nothing in ember data is iterable

Do you mean you can't do `for (const treatment of versionedTreatments)` because `versionedTreatments` is not iterable? I _think_ you can just call [`toArray()`](https://api.emberjs.com/ember/3.26/classes/Ember.NativeArray/methods/toArray?anchor=toArray) on to it first?

Anyway, I still think the real solution is just converting everything to sync relationships since that will make things a lot more straightforward. I really hoped they would have made that the default in 4.0 since all the core members seem to agree on that at least. I heard you like going on refactoring sprees? :grimacing:

Windvis: I also think it's ok to leave it like this since it works. We can always deal with the deprecations later if needed.

If you want to be really sure you could always post in the #ember-data channel in the Discord. Runspired is very active there and will certainly have the answer for you.

app/services/publish.js
```diff
@@ -0,0 +1,106 @@
...
+  get treatmentExtracts() {
+    return [...this.treatmentExtractsMap.values()].sortBy('position');
+  }
+
+  get isIdle() {
+    return this.loadExtractsTask.isIdle;
+  }
+
+  get loadExtractsTask() {
+    return this._loadExtractsTask.unlinked();
```
nvdk, line 37: I'll have to read up on what unlinked means here

nvdk (changes requested):
This looks quite nice actually, I was just considering whether it wouldn't be better if the backend handled the preview vs persisted view for us. With the service this is nicely abstracted away in case we do ever change that.

Wrt
> Issues can probably appear since this is basically doing ad-hoc caching, depending on the developer to reload the cache. I don't know if this is a good pattern or not. It seems to be in the spirit of what ember-storefront proposes, but I might be wrong about that.

perhaps it's a good idea to add a comment to that effect in the service?

I did notice we no longer do pagination in the service though, so that needs to be fixed I think?

app/controllers/meetings/publish/uittreksels.js
```diff
@@ -1,92 +1,75 @@
...
   @task
-  *fetchUittreksels() {
-    const uittreksels = [];
-    const previews = yield this.fetchExtractPreviews.perform();
-    for(const uittreksel of previews) {
-      const existingUittreksels = yield this.store.query('versioned-behandeling',{
-        'filter[behandeling][:id:]': uittreksel.data.attributes.uuid,
-        include: 'signed-resources,published-resource,behandeling'
-      });
-      if(existingUittreksels.length) {
```
nvdk, line 45: this piece is missing in the service version!

app/services/publish.js
```diff
@@ -0,0 +1,106 @@
...
+   * id of the treatment
+   * @param meetingId
+   * @return {Generator<*, void, *>}
+   */
+  @task
+  *_loadExtractsTask(meetingId) {
+    const [newExtracts, meeting, treatments, versionedTreatments] = yield all([
+      fetch(`/prepublish/behandelingen/${meetingId}`).then((res) => res.json()),
+      this.store.findRecord('zitting', meetingId),
+      this.store.query('behandeling-van-agendapunt', {
```
nvdk, line 51: this will only fetch the first 20

app/controllers/meetings/publish/uittreksels.js
```diff
@@ -1,92 +1,75 @@
...
 
   get meeting() {
     return this.model;
   }
 
-  @task
-  * initializeUittreksels() {
-    yield this.fetchUittreksels.perform();
+  get extracts() {
+    return this.publish.treatmentExtracts;
```
nvdk, line 26: I don't think this is sorted

abeforgit: It's just sorted in the service already, frontend sort

app/services/publish.js
abeforgit: right always forget this, good catch

app/services/publish.js
abeforgit: it is the stupidest thing about ember-concurrency
https://github.com/machty/ember-concurrency/issues/161
https://github.com/machty/ember-concurrency/issues/249
You get a warning in the console at least. But it is really annoying, especially since you can pass an unlinked task through like here, but if it needs to be linked, you can't do that since you have to do it right before a perform or something

app/services/publish.js
abeforgit: Yeah I think at least in GN going all sync would have a lot of value
I get what the async relations are going for but the syntax looks way too much like normal js on the one hand, and is annoying on the other (like in this case, you always have to save every relationship to a variable, you cant use the already existing model object)

app/services/publish.js
abeforgit: fun fact: this was always the case

app/services/publish.js
abeforgit: gonna leave it, too close to release to muck about

abeforgit:
added comments & did pagination
I think it would be better to get rid of the state in the service and provide a way for the print route to simply get what it needs based on the id of the treatment
service can then serve as a nice wrapper around the publishing flow, abstracting away the fetch calls and stuff but otherwise just exposing tasks and not managing any of its own state

but for now, this'll do I think

app/services/publish.js
```diff
@@ -51,10 +62,12 @@ export default class PublishService extends Service {
       this.store.query('behandeling-van-agendapunt', {
         'filter[onderwerp][zitting][:id:]': meetingId,
         include: 'onderwerp',
+        page: { size: 1000 },
```
nvdk, line 65: this will do for now, but requesting page per page would be my recommended solution to avoid request timeouts

abeforgit: yeah I know
we really need to abstract that away at some point
it's way too much of a pain to add every time

PR #135, opened 2021-06-11
Bugfix/participants sync

abeforgit:
fixes https://binnenland.atlassian.net/browse/GN-2677

Hopefully for the last time. Introduces `ember-could-get-used-to-this` as a dependency for the use of the Resource paradigm
([more info](https://www.pzuraq.com/introducing-use/)) 

I think this had quite a few more bugs than we thought. The original code was so bad I'm surprised it even did anything at all (I can be this harsh, cause it was my code :joy: )

Honestly, without resources, this would've been an absolute mess of a flow to implement. The pattern of "default from arguments, but allow local changes, but when arguments change, also recalculate local state" is really streamlined by them. Also no cumbersome `did-insert` needed, all the relevant update logic is now clearly visible in the backing class alone.

The implementation of the Resources used here can definitely use some work still. You have to be very careful about reading tracked state in the lifecycle methods, this can easily lead to weird problems (and this is counterintuitive, cause _reading_ is usually never something you expect to have side-effects). [This thread](https://github.com/pzuraq/ember-could-get-used-to-this/pull/37#issuecomment-814157755) is highly recommended reading on this topic.


app/components/participation-list/mandataris-row.hbs
```diff
@@ -4,7 +4,7 @@
   <td>{{moment-format @mandataris.isBestuurlijkeAliasVan.geboorte.datum 'DD/MM/YYYY'}}</td>
   <td>
     <label class="checkbox checkbox--block">
-      <Input class="checkbox__toggle" @type="checkbox" @checked={{this.selected}} @click={{this.toggle}} />
+      <input class="checkbox__toggle" type="checkbox" checked={{@selected}} {{on "input" (fn @toggle @mandataris @selected)}} />
```
nvdk, line 7: I don't really get why you pass selected here, is that so they have the previous value?

abeforgit: No the selected state is fully managed in the  parent components, so this is the latest state
Note it went from Input (2-way binding) to input (1-way)

abeforgit: I just forgot to remove the now defunct backing class

nvdk: but `@selected` wont be updated here yet right, because it's 1 way binding.  the actual value after change/input would be in event.target.selected (event will be curried into your function I believe). given it's a toggle function I was guessing the value wasn't used (and thus shouldn't be passed?), but if it is used I think it might do the inverse of what you want :).

Sidenote: I prefer the change event for checboxes but I think both input and change would work here.

nvdk: having had a look at the toggleParticipant method, it will work because you expect the previous and not the actual value to be passed. ok that makes sense, though not entirely what I'd expect.

nvdk (changes requested):
Almost everything works perfectly :partying_face: 
... except rollback when closing an edit modal. the changes are not rolled back and shown in the meeting overview. this might already have been an issue. in the mean confirmed this was not the case before

app/components/participation-list/mandataris-row.hbs
abeforgit: I still dont quite understand what you mean
Though originally I didnt pass the selected value up to the handler (cause I could just get it there from the map)
but as I was debugging another issue, this was one of the things that changed
didnt look out of place so I left it I suppose

abeforgit:
it never ends, ill have a look

app/components/participation-list/mandataris-row.hbs
nvdk: for a callback from an input field (which is what this is) I typically expect to receive the actual value of that input. in this case you pass the value before the input. 

- you pass `@selected=false` down to the mandataris-row. the checkbox has checked=false
- user clicks the checkbox, the checkbox is now checked
- the component calls the callback with (mandataris, false)
- you negate this false in your toggle function

So what I'd expect (but this is nitpicking) to see is either not passing the value when toggling (the parent has the old value already after all) or passing the new value (true in my example)

abeforgit:
I still don't know how the original code even worked
like the map for the participants wasn't even tracked
weird stuff

abeforgit:
just found another bug, in a new meeting the meeting participants default to absent

PR #136, opened 2021-06-14
Feature/limit bestuursorgaan

abeforgit:
fixes https://binnenland.atlassian.net/browse/GN-2734

Filter out admin bodies with an endDate older than two months before the current date

### DIFF
--- a/app/components/administrative-body-select.js
+++ b/app/components/administrative-body-select.js
@@ -2,6 +2,7 @@ import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { task } from 'ember-concurrency';
+import moment from 'moment';
 
 const VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS = [
   "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000005", //	"Gemeenteraad"
@@ -27,7 +28,6 @@ const VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS = [
  * @property {string} id Input id so it can be linked with a label
  * @property {BestuursOrgaan} selected which governing body is currently selected
  * @property {(administrativeBody: BestuursOrgaan) => void} onChange change handler called when a body is selected
- * @property {(endDate: string) => boolean} endDateFilter filter function to restrict shown bodies by their end date
  * @property {boolean} error whether there is a form value error and we should render as such
  */
 
@@ -43,6 +43,10 @@ export default class AdministrativeBodySelectComponent extends Component {
     this.fetchAdministrativeBodies.perform();
   }
 
+  /**
+   * Fetch bodies which are of the right classification, and whose
+   * end date is not older than 2 months before the current date
+   */
   @task
   * fetchAdministrativeBodies() {
     let currentAdministrativeUnitId = this.currentSession.group.id;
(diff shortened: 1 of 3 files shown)


app/components/administrative-body-select.js
```diff
@@ -45,9 +56,24 @@ export default class AdministrativeBodySelectComponent extends Component {
...
+        );
+        const bodyIsValid =
+          VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS.includes(classificationUrl);
+        // shortcutting to avoid work
+        if (!bodyIsValid) {
+          return false;
+        }
+
+        const endDate = administrativeBodyInTime.bindingEinde;
+        if(this.args.endDateFilter) {
```
Windvis, line 72: This seems oddly specific and a bit of a weird component design tbh. Is there a reason why this can't be part of the component? This seems like a premature "keep things generic" optimization. I think this component is only used once, right?

app/components/administrative-body-select.js
```diff
@@ -21,6 +21,17 @@ const VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS = [
   "http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000008" //	"Vast Bureau"
 ];
 
+/**
+ * @typedef {Object} Args
+ *
+ * @property {string} id Input id so it can be linked with a label
```
Windvis, line 27: What does `@property` do exactly? What does TS do with this annotation?

Windvis:
Woops, didn't notice that you removed my review request :grimacing:.

app/components/administrative-body-select.js
nvdk: I agree, I think even if it's reused it's very likely the same filter will be used.

abeforgit:
> Woops, didn't notice that you removed my review request grimacing.

your reviews are always welcome, just didn't want to overload you :)

app/components/administrative-body-select.js
abeforgit: fair enough, just thought it might be used somewhere else 
but even then the current design is odd, if we want a configurable filter it should receive the whole mandatee

app/components/administrative-body-select.js
abeforgit: This whole typedef is just the syntax for defining an object in jsdoc, as seen [here](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#typedef-callback-and-param)

We then document the component by saying it extends `Component<Args>` which is how glimmer components are typed. In theory, with a compliant typescript service, this should allow your editor to provide type hinting and autocompletion on the `this.args` object. In practice, this works quite well in emacs and vscode, not so great in webstorm. 
You could even setup the ts compiler to really typecheck these and error out when it's not right, but given we have such little type information in the project I think that's not a great idea.

So its usefulness for ts is a bit of a mixed bag. But since we need to document the arguments for a component _anyway_ (cause you know, thats the interface of the thing) I figured why not use the syntax that can actually be used by tooling.

You see im not going all the way here by not importing the `BestuursOrgaan` model type, cause that becomes cumbersome very quickly (the syntax is hideous) and provides no extra benefit since our models are also not typed/documented.

nvdk (approved):

app/components/administrative-body-select.js
Windvis: Ah, thanks for the info! The "property" part made me think it was related to the class somehow. I'm not familiar with adding types through comments, as evidenced :smile:. Did you take a look at the [glint](https://www.salsify.com/blog/engineering/introducing-glint-a-typed-template-solution-for-glimmerx-and-ember) project btw? If not I think you will like it :grimacing:.

Windvis: No, not really, no. But it's cool that something like that is in the works.

PR #137, opened 2021-06-22
bugfix: provide read only view, so that once the agendapoint is published the voting is visible in the meeting

lagartoverde:


app/components/behandeling-van-agendapunt.hbs
```diff
@@ -65,11 +65,11 @@
 {{/if}}
 
 
-{{#if (and this.hasParticipants this.editable)}}
+{{#if (and this.hasParticipants)}}
```
nvdk, line 68: no need for the and here

PR #138, opened 2021-06-23
Bugfix/store public setting of treatment

nvdk:
turns out we didn't actually save this :facepalm: . ty @Asergey91 for mentioning!


Asergey91:
doesnt seem to work atm 
something to do with the getters
this.openbaar returns undefined
while this.args.behandeling.openbaar returns the correct value

abeforgit:
You forgot to remove `@tracked openbaar` from the class properties when converting to a getter
classic mistake

abeforgit (changes requested):
You forgot to remove `@tracked` openbaar from the class properties when converting to a getter
classic mistake

works well once you remove that

PR #139, opened 2021-06-23
Allow governor to be selected if deputation

abeforgit:
solves https://binnenland.atlassian.net/browse/GN-2747

Afaict only vlaams-brabant actually has data for this atm, so it is only tested for them.


lagartoverde (approved):
Looks good and now it works 👍

PR #140, opened 2021-06-23
Feature/get draft description

lagartoverde:
Get description of the draft agendapoint when selecting it

### DIFF
--- a/app/components/agenda-manager/agenda-item-form/select-draft.js
+++ b/app/components/agenda-manager/agenda-item-form/select-draft.js
@@ -2,7 +2,7 @@ import Component from '@glimmer/component';
 import { inject as service } from "@ember/service";
 import { action } from "@ember/object";
 import { tracked } from "@glimmer/tracking";
-import { restartableTask } from "ember-concurrency";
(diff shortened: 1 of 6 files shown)


lagartoverde:
Sorry, I converted to a draft because I reread the issue and it should be a button not automatic

lagartoverde:
Ready for review now :)

Asergey91:
hm couldn't get it to work with a fresh document with only the description filled in and nothing else.

lagartoverde:
Realized that this was not consistent because it was missing the prefixes for marawa to work 100% correct, now it should work every time

nvdk (approved):
looks good and works ok! 

if there's no description the behaviour feels a bit wonky, in that the user doesn't have any feedback what so ever. 

perhaps some kind of message if we don't find a description would be nice, or a loading bar while processing so users know something is happening?

lagartoverde:
Sorry, I forgot about this, I tried adding a loading bar but the function returns instantly, maybe a message would be nice, but not sure where to put it, @Dietr do you have something to say about this? :)

PR #146, opened 2021-07-02
updated app universum and changed from au button to au link

lagartoverde:
Waiting for https://github.com/appuniversum/ember-appuniversum/issues/120


nvdk (changes requested):
I think it would be best to remove the goToEdit action as well in the component.
A similar change can also be done for the editIntro and editOutro in the meeting form

PR #151, opened 2021-07-09
Fixed behandeling publishing status was not being updated properly without page refresh

lagartoverde:


app/components/behandeling-van-agendapunt.js
```diff
@@ -100,13 +100,16 @@ export default class BehandelingVanAgendapuntComponent extends Component {
...
-        if (status.id === PUBLISHED_STATUS_ID) {
-          this.published = true;
-        }
+    const behandeling=(yield this.store.query("behandeling-van-agendapunt", {
+      "filter[:id:]": this.args.behandeling.id,
+      include: 'document-container.status'
+    })).firstObject;
+
+    if(behandeling){
+      const statusId = behandeling.get('documentContainer.status.id');
```
abeforgit, line 109: I know this works, but as far as my understanding of ember data goes, it's not guaranteed to work
this is better (again, afaik):
```js
      const status = yield behandeling.get('documentContainer.status');
      const statusId = status.id;
```
You may think that this will generate an extra network request, but it doesn't, cause ember-data is smart enough to know it can use the cached value 
(smart is maybe the wrong word, I think it might be just a timeout)
In any case, with an extra yield you are sure it's resolved, and you also dont make any extra network request when it's not needed

And you should not do
```js
const statusId = yield behandeling.get('documentContainer.status.id')
```
this `get` function isn't what you (or at least I) think intuitively. It doesnt magically promisify everything, it literally just gets the property, with some proxy-awareness. So if you get `documentContainer.status.id`, it's not gonna give you a promise, only a proxy (if the relationships are not resolved) cause `id` is not a relationship and thus not a promise.

Again, because you use include, this will actually work because everything is already resolved. But again, it's not _guaranteed_ to and using this pattern in other situations has bit me in the past

abeforgit (changes requested):
It works, some comments about the code

abeforgit (approved):

PR #152, opened 2021-07-12
Feature/file upload

Asergey91:
further work on attachments

### DIFF
--- a/app/controllers/agendapoints/edit.js
+++ b/app/controllers/agendapoints/edit.js
@@ -13,6 +13,9 @@ export default class AgendapointsEditController extends Controller {
   @tracked displayDeleteModal = false;
   @tracked _editorDocument;
   profile = 'draftDecisionsProfile';
+  @tracked uploading=false;
+  @tracked decisions = [];
+  @service documentService;
 
   get editorDocument() {
     return this._editorDocument || this.model.editorDocument;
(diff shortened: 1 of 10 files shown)


nvdk:
@lagartoverde : some missing semi's in app/services/document-service.js

lagartoverde:
Oh, I didn't check the semis yet, this is not ready, the function to get the decisions only runs when opening the modal, I will solve that next now that I finished the prepublisher

lagartoverde:
Ready for review now :)

abeforgit (commented):
Looks ok from what I understand of it. Seems to work as well.

May or may not be an issue:
when adding attachments from the draft-decision screen, the "besluit" select box is always empty, even when it does show a value in the zitting view.

app/controllers/agendapoints/edit.js
```diff
@@ -13,6 +13,9 @@ export default class AgendapointsEditController extends Controller {
   @tracked displayDeleteModal = false;
   @tracked _editorDocument;
   profile = 'draftDecisionsProfile';
+  @tracked uploading=false;
```
nvdk, line 16: please mind the formatting

app/controllers/meetings/edit/treatment.js
```diff
@@ -14,6 +14,9 @@ export default class MeetingsEditTreatmentController extends Controller {
   @tracked documentContainer;
   @tracked document;
   @tracked initialContent;
+  @tracked uploading=false;
```
nvdk, line 17: formatting

app/services/document-service.js
```diff
@@ -32,4 +32,24 @@ export default class DocumentService extends Service {
     }
     return prefixesString;
   }
+  getDecisions(editorDocument) {
+    const node = document.createElement('body');
```
nvdk, line 36: this shares quite a bit of logic with getDescription which can be abstracted

nvdk (changes requested):
- Some small remarks on formatting
- I'd recommend abstracting the extractDecisions part that's present in both getDescription and getDecisions in the document service
- attachment type "regelgevend" is only possible for an attachment linked to a decision

nvdk (approved):

PR #157, opened 2021-08-12
Feature/file upload

Asergey91:


Dietr (approved):
I added some changes:
- made the attachment/document button into tab buttons to improve accessibility and clarity
- replaced some new css with available components
- I noticed the data table search was broken so added a quickfix for that (a bit out of scope of this PR but seemed a bit overkill to create a new one for that)

translations/nl-BE.yaml
```diff
@@ -347,4 +347,4 @@ attachments:
   type: Type
   decision: Besluit
   typePlaceholder: geen
-  regulatory: regelgevend
\ No newline at end of file
+  regulatory: regelgevend
```
abeforgit, line 350: make sure to add this one to the english translation file as well

abeforgit (changes requested):
When you add a decision to an agenda-item, you need to save the document first before it shows up in the besluit dropdown. There's no easy way around this, so we should at least make it clear to the users this is what they have to do.
I think the best option is to save automatically when going to bijlagen, with a confirm popup asking if they are sure they want to save.

We can also rename the "bijlagen" button to "opslaan en bijlagen toevoegen", eliminating the need for the confirm popup.

translations/nl-BE.yaml
Asergey91: done

PR #158, opened 2021-08-12
Fix print view overflow issue when &nbsp is used in text

Dietr:
The use of &nbsp; in text means that this is handled as one long word and won't break with the width of the container.

Added:
- Quick fix with CSS overriding the behaviour of &nbsp in long text. 

Possible better solution in the future:
- remove &nbsp; when a user copies text in the editor

### DIFF
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -416,3 +416,10 @@
     flex-wrap: nowrap !important;
   }
 }
+
+// FIX NBSP ISSUE
+.say-content,
+.au-c-editor-preview {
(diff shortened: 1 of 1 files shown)


abeforgit:
I was actually looking for a non-collapsing, but breaking space when developing the editor, unfortunately that doesn't exist, adding extra logic to space handling. We do actually want breaking everywhere, its just the whitespace collapsing that's never desired, so this seems like a good workaround.
I am wary of changing default rendering behavior though, so I've added a story to investigate the options further:
https://binnenland.atlassian.net/browse/GN-2868

PR #159, opened 2021-08-13
Feature/delete meetings

Asergey91:
backend pr: https://github.com/lblod/app-gelinkt-notuleren/pull/39
meeting-service repo: https://github.com/lblod/meeting-service/


lagartoverde:
Code looks good and seems to work, but waiting for further testing when the backend branch is ready, so I can test with the final setup

nvdk:
currently hanging because of a (possible) cache issue: it seems the cache doesn't update correctly after removing a meeting

nvdk:
> currently hanging because of a (possible) cache issue: it seems the cache doesn't update correctly after removing a meeting

this should resolved with mu-cl-resource >= 1.21.0

PR #160, opened 2021-08-16
Consume job for prepublishing

madnificent:
The prepublisher emits jobs for lengthy operations.  With this we consume these in a polling loop.

Next:
- [ ] We may want to use this for the notulen too

### DIFF
--- a/app/services/publish.js
+++ b/app/services/publish.js
@@ -57,23 +57,7 @@ export default class PublishService extends Service {
   @task
   *_loadExtractsTask(meetingId) {
     const [newExtracts, meeting, treatments, versionedTreatments] = yield all([
-      async () => {
-        let uuidResp = await fetch(`/prepublish/behandelingen/${meetingId}`);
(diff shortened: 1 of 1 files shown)


abeforgit:
While this works now with the small changes I did, it is _really_ slow (about 3 times slower on big queries, going from a painful but workable 2 minutes to 6-7 minutes on a local setup)
At that rate we might exceed our "surely long enough" 10 minute timeout sooner than we thought.

madnificent:
Are you suggesting this commit makes the operation slower?  I would be surprised if this would slow down the frontend or backend side considerably versus the old implementation.  Aside from not timing out, that is.

abeforgit:
I will do more testing, but yes, that's what seemed to happen
We are also exploring other avenues, but this solution will probably still be needed anyway (mostly for the notulen)

For the uittreksels, we can actually defer generating the preview 1 step in the UI flow, which means we can get away with only generating the preview for the uittreksel the user is interested in, which is both a lot easier on the server and nicer for the user (no big upfront loading cost)

PR #163, opened 2021-08-23
lazy load uittreksels

nvdk:
this is a somewhat quick hack and isnt the prettiest, but it should do what it's
supposed to do: allow users to only load the data the need and work with that.

Currently still has some rendering issues to iron out.


lagartoverde (approved):
Looks good, some remarks I can fix if you want:
- Doesn't seem obvious at first that you have to click the name of the agendapoint in order to publish it, maybe we should add a button on the table or something to make it clearer
- The searching works but takes some time, maybe we should have a loading icon or something
- Agendapoints are 0 based on the index, I think they should be 1 based
- There's a dot on top of the publish screen, due to the <li> container

Dietr (approved):

app/routes/meetings/publish/uittreksels/show.js
```diff
@@ -0,0 +1,13 @@
+import Route from '@ember/routing/route';
+
+export default class MeetingsPublishUittrekselsShowRoute extends Route {
+  async model(params) {
+    console.log(params);
```
nvdk, line 5: stray debugging statement should go

PR #167, opened 2021-08-27
Use task flow for preview, signing and publication of notulen

abeforgit:
depends on https://github.com/lblod/notulen-prepublish-service/pull/61

### DIFF
--- a/app/services/mu-task.js
+++ b/app/services/mu-task.js
@@ -22,7 +22,7 @@ export default class MuTaskService extends Service {
   @task
   * waitForMuTaskTask(taskId, pollDelayMs = 1000, timeoutMs = 300000) {
     let resp;
-    let currentStatus
+    let currentStatus;
     const startTime = Date.now();
     do {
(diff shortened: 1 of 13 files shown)


nvdk (approved):
we will have to move the decision signing/publication to these abstractions as well.

lagartoverde (approved):

app/services/mu-task.js
```diff
@@ -0,0 +1,55 @@
...
+      resp = yield this.fetchMuTask(taskId);
+      if(resp.ok) {
+        currentStatus = (yield resp.json()).data.attributes.status;
+      } else {
+        currentStatus = null;
+      }
+
+    } while(resp.ok
+            && currentStatus === TASK_STATUS_RUNNING
+            && Date.now() - startTime < timeoutMs);
```
nvdk, line 37: general remark: I personally prefer a number of tries over a timed timeout. on a slow connection, with a lot going on this might only try very few times.

PR #169, opened 2021-09-01
Make editor feature flags configurable from environment

abeforgit:


nvdk:
as discussed we're no longer pursuing this

PR #171, opened 2021-09-02
added a toggle to toggle all agendapoints to be public/private in notulen preview

lagartoverde:


lagartoverde:
Screen Shot 2021-09-02 at 17 12 28
@Dietr it currently looks like this, maybe you want to add some styling to it :)

Dietr:
@lagartoverde Is it possible to use a checkbox like in the toggle component (https://appuniversum.github.io/ember-appuniversum/docs/atoms/au-toggle-switch)?

lagartoverde:
Is it possible, but I think is not really what we want, as we already have checkboxes for individual behandelings I think another checkbox for all of them is going to be a little bit confusing

nvdk:
I could see a toggle at the top working, @lagartoverde can you change it to the toggle so @Dietr can evaluate?

lagartoverde:
Done :)

Dietr (approved):

PR #173, opened 2021-09-06
use au-file-upload from appuniversum

nvdk:
depends on https://github.com/appuniversum/ember-appuniversum/pull/139 , so
needs a version bump of appuniversum before it can be merged


nvdk:
woops, foobarred the branch. will create new PR soon

PR #174, opened 2021-09-07
Add styling for tables still relying on webuniversum css

Dietr:
Some tables were still relying on webuniversum. Added css for these cases and removed the old webuniversum classes.


lagartoverde (approved):
Good job 👍

PR #176, opened 2021-09-08
Hide attachments behind feature flag

lagartoverde:
I think we only need to hide the button as that would make the user unable to upload any attachment


nvdk:
probably best to add the flag in environment.js?

abeforgit (changes requested):
you forgot to hide the button in the agendapoint edit route

PR #177, opened 2021-09-08
Calculated status based on the actual publication status of the extract

lagartoverde:
Depends on https://github.com/lblod/app-gelinkt-notuleren/pull/45

### DIFF
--- a/app/components/signatures/publication-status.js
+++ b/app/components/signatures/publication-status.js
@@ -10,7 +10,7 @@ export default class SignaturesPublicationStatus extends Component {
   @action  
   async loadResource() {
     console.log(this.args.versionedResource.get('signedResources'));
-    this.signedResources = await this.args.versionedResource.get('signedResources');
+    this.signedResources = await this.args.versionedResource.get('signedResources') || [];
     this.publishedResource = await this.args.versionedResource.get('publishedResource');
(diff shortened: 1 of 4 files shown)


nvdk:
I don't think this will work as expected, as it doesn't await the load of signedResources?

lagartoverde:
You are totally right, I trusted a stackoverflow comment and then tested a little bit and it seemed to work ok, but testing more in depth shows signedResource didn't work as expected

app/components/signatures/publication-status.js
```diff
@@ -0,0 +1,28 @@
...
+
+
+export default class SignaturesPublicationStatus extends Component {
+  @tracked signedResources = [];
+  @tracked publishedResource
+  @tracked ready = false;
+  @action  
+  async loadResource() {
+    console.log(this.args.versionedResource.get('signedResources'));
+    this.signedResources = await this.args.versionedResource.get('signedResources');
```
abeforgit, line 13: you are overwriting the empty array here with something that is potentially null or undefined(if there are no signedResources), and then your publicationStatus logic crashes cause you are trying to access the length of it.
so either add a `|| []` here or change the logic in the getter

abeforgit (changes requested):

PR #178, opened 2021-09-08
Fix notulen was not correctly reloaded under a very specific scenario

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-2902

### DIFF
--- a/app/controllers/meetings/publish/notulen.js
+++ b/app/controllers/meetings/publish/notulen.js
@@ -37,14 +37,16 @@ export default class MeetingsPublishNotulenController extends Controller {
       yield Promise.all(versionedNotulens.map(async (notulen) => {
         const publishedResource = await notulen.publishedResource;
         const signedResources = await notulen.signedResources;
+        let notulenSet = false;
         if(publishedResource) {
           this.publishedResource = publishedResource;
           this.publicBehandelingUris = notulen.publicBehandelingen || [];
           this.notulen = notulen;
+          notulenSet = true;
         }
         if(signedResources.length) {
           this.signedResources = signedResources;
-          if(!this.notulen) {
+          if(!notulenSet) {
             this.notulen = notulen;
           }
         }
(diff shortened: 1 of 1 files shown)


app/controllers/meetings/publish/notulen.js
```diff
@@ -37,14 +37,16 @@ export default class MeetingsPublishNotulenController extends Controller {
       yield Promise.all(versionedNotulens.map(async (notulen) => {
```
abeforgit, line 37: we map over versionedNotulen but on every iteration all we do is overwrite the local state.
as I understand it we could then just do the logic on the last element of the array no?

not even that, it uses Promise.all which means its arbitrary which will be the final state, no?

app/controllers/meetings/publish/notulen.js
```diff
@@ -37,14 +37,16 @@ export default class MeetingsPublishNotulenController extends Controller {
...
         const publishedResource = await notulen.publishedResource;
         const signedResources = await notulen.signedResources;
+        let notulenSet = false;
         if(publishedResource) {
           this.publishedResource = publishedResource;
           this.publicBehandelingUris = notulen.publicBehandelingen || [];
           this.notulen = notulen;
+          notulenSet = true;
```
abeforgit, line 45: why aren't we simply always setting this.notulen?

app/controllers/meetings/publish/notulen.js
lagartoverde: No, because there will be 2 versioned notulens 1 with signedResources and one with publishedResources

lagartoverde: We want to get both

app/controllers/meetings/publish/notulen.js
lagartoverde: We only set the notulen if a previous version with a published resource hasn't already set it.
This comment made me realize the place of the notulenSet declaration was not correct I changed it

app/controllers/meetings/publish/notulen.js
abeforgit: ah I see, fair enough

abeforgit:
I can still reproduce the bug with the latest code...

PR #179, opened 2021-09-09
[HOTFIX] Add updated items to changeset

abeforgit:
When not updating the position, the agenda item did not get added to the changeset, thus edits like changing a title were not saved.


nvdk:
I've rebased this on v2.11.0 as that's the version on prod at the moment

PR #182, opened 2021-09-13
added agendapoint position to the select and the search

lagartoverde:
I know in the planning we said to leave the search functionality for later, but I took a look at the powerselect component and it seemed kinda easy to do. Let me know if I'm missing something but it worked when I tested it


nvdk (changes requested):
not sure if related to this PR, but the positions shown are incorrect after making a selection

nvdk:
in any case this makes it very visible, so best to fix it before we merge this :)

nvdk (approved):

PR #185, opened 2021-09-14
use au-file-upload from appuniversum

nvdk:


Asergey91 (approved):

Asergey91:
tested, seems to work as expected

PR #190, opened 2021-09-15
added attachment counter to ap edit route, fixed bug with undefined length

Asergey91:
found this while reviewing au-file upload pr


nvdk (approved):
good find

PR #191, opened 2021-09-15
improved flow for creating agendapoints

nvdk:
Some thoughts:
- doesn't include the selection of the decision type yet, perhaps best done in another PR
- integrates parts of the template plugin, so must make sure to keep the code in sync when updating that plugin
- probably need to provide something similar in the agendapoint manager in the meeting itself


translations/nl-BE.yaml
```diff
@@ -359,3 +363,7 @@ attachments:
   typePlaceholder: geen
   regulatory: regelgevend
   saveAndGo: Opslaan en bijlagen toevoegen
+documentCreator:
```
lagartoverde, line 366: These strings are missing on the English version

lagartoverde (approved):
Works good, only some translations are missing, good job 👍

translations/nl-BE.yaml
```diff
@@ -25,7 +25,11 @@ inbox:
     createdOnLabel: Aangemaakt
     updatedOnLabel: Aangepast
     statusLabel: Status
-    new: Nieuw agendapunt
+    new:
+      title: Nieuw agendapunt
+      errorSaving: Er liep iets fout bij het aanmaken van het agendapunt
+      create: Opslaan
```
Dietr, line 31: I would call this "aanmaken"/"create" as it is a new agenda item

Dietr (approved):

PR #194, opened 2021-09-17
add brackets around planned label

Dietr:
The "planned" label was not styled and caused confusion when displayed after a title in the publication view.
Adding brackets separates this from the sentance and makes he meaning clear.


nvdk (commented):
Should we just do this in the template of the publication?

Dietr:
Yes, that's also an option. Can you tell me where I can find this template?

nvdk:
@Dietr you can find it here: https://github.com/lblod/notulen-prepublish-service/blob/4061fd1859194c859665918d643e6d4a84feda04/support/templates/partials/treatment.hbs#L12

The benefit of doing it there is that it will also be there if it's ever shown without styling, the downside being that already published documents will not have the fix.

PR #197, opened 2021-09-20
added an overview of attachments in meeting interface

Asergey91:
gn 2908


Asergey91:
is there a float:right class in app universum? couldnt find it

Asergey91:
https://github.com/lblod/frontend-gelinkt-notuleren/pull/197

Dietr (approved):
Refactored the title component so we don't need a float class. We prefer to use flexbox for positioning.

Also noticed we were not using the appuniversum file upload styles so also fixed that.

nvdk (approved):

PR #198, opened 2021-09-21
Fix issue with double characters

abeforgit:
As far as I can tell, this was caused by the besluittype plugin generating errors which caused some rendering hiccups. The latest version of that plugin solves the issue. 

The problem only appeared on prod, cause the kind of errors it generated are only warnings when running local dev server, but errors on prod. 

This pr is then just a version bump, but also a bugfix

Fixes https://binnenland.atlassian.net/browse/GN-2944


abeforgit:
If we want to hotfix release this, here's a reminder of the flow:
https://abb-vlaanderen.gitbook.io/feature-passports-lblod/gn/release-guide-gn#bugfix-aka-patch-or-hotfix-releases

PR #200, opened 2021-10-03
Internal/ember upgrade

nvdk:
upgrading to ember 3.28 triggers a lot of deprecations (and some bugs in our code). tracking the progress here.


Dietr:
Adding labels which are linked to the input field with an id should fix this. If the label is not necessary as a visual aid you can add the `au-u-hidden-visually` class.

Windvis:
Heads up: There seem to be some issues in the 3.28 release of Ember Data which probably affect our projects: 

- https://github.com/emberjs/data/issues/7684
- https://github.com/emberjs/data/issues/7688

Pinning the ember-data package to 3.27 might probably a good idea until those are resolved.

package.json
```diff
@@ -26,91 +26,102 @@
...
+    "ember-cli": "~3.28.0",
+    "ember-cli-app-version": "^5.0.0",
+    "ember-cli-autoprefixer": "^2.0.0",
     "ember-cli-babel": "^7.26.11",
     "ember-cli-dependency-checker": "^3.2.0",
     "ember-cli-dependency-lint": "^2.0.0",
-    "ember-cli-htmlbars": "^5.7.2",
-    "ember-cli-inject-live-reload": "^2.0.2",
-    "ember-cli-sass": "^7.1.4",
+    "ember-cli-deploy": "^1.0.2",
```
nvdk, line 66: you can drop the ember-cli-deploy related dependencies

ember-cli-build.js
```diff
@@ -18,29 +22,22 @@ module.exports = function (defaults) {
...
+      sourceMap: !envIsProduction,
+      sourceMapEmbed: !envIsProduction,
     },
     autoprefixer: {
       enabled: true,
       cascade: true,
-      sourcemap: true
+      sourcemap: !envIsProduction
     },
     flatpickr: {
```
nvdk, line 34: I guess flatpickr config can go now as well

benjay10: Updated the citaten-plugin version with AppUniversum component.

Windvis:
While trying to update the loket app I also ran into the same Ember Data issue that was mentioned in this commit: https://github.com/lblod/frontend-gelinkt-notuleren/pull/200/commits/9ea5695e7c772373993d9bd38efeb30ff6ea7dc8

I did some digging and in our case it was related to async relationships.

The setup was something like this:
- Component X receives a record as an arg
- in the constructor a hasMany relationship is read, and set to a local property on the component: 
  `this.property = this.args.record.get('relationshipName');`
- When saving this property is set on the record again:
  `this.args.record.set('relationshipName', this.property)`

The issue is that the relationship promise is never resolved so the PromiseManyArray is stored instead of the actual relationship. When setting that PromiseManyArray to the relationship again, the error is thrown that an array is needed since it's not actually one. 
I think this is still a real bug in Ember Data but it's a bit of a gray area since the array methods on the PromiseManyArray [are open for deprecation](https://github.com/runspired/rfcs/blob/ember-data/deprecate-model-static-field-access-without-lookup/text/0741-ember-data-deprecate-model-static-field-access-without-lookup.md) and working with it directly has always been a bad practice anyways. I think it usually happens by accident because everything works.

So if the situation here is similar, the real fix is to make sure relationships are fully resolved before working with them. Also, I think 3.28.8 also included a bug fix for Ember Data, so that might also fix other issues were encountered here.

PR #204, opened 2021-10-06
Only save document if it's in a dirty state

lagartoverde:
I used the dirty attribute we are using to show the confirmation modal when we are leaving the route


nvdk:
we will also need this on the agendapoint page

abeforgit:
ah wel uh
I also approve these changes :joy:

PR #205, opened 2021-10-06
add a warning to attachment overview if no decisions were found

nvdk:
warningexample
An attempt to make it a bit more clear how it works. 

@Dietr not sure about the placement of the warning (currently inside the table and feels a bit messy). 
In general: might consider just converting this to a regular info box and always show it?

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -364,6 +364,8 @@ attachments:
   typePlaceholder: none
   regulatory: regulatory
   saveAndGo: Save and go to attachments
+  noDecisonsFound: No decisions found
+  decisionRequired: Only attachments linked to a decision are included in the publication.
 documentCreator:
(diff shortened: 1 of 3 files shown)


translations/en-us.yaml
```diff
@@ -364,6 +364,8 @@ attachments:
   typePlaceholder: none
   regulatory: regulatory
   saveAndGo: Save and go to attachments
+  noDecisonsFound: No decisions found
```
lagartoverde, line 367: Typo here :) `noDecisionsFound`

lagartoverde (changes requested):
Looks good but there's a typo ;)

Dietr (approved):
I also prefer not to have this inside the table so changed the position of the warning to the bottom of the screen.

PR #206, opened 2021-10-06
convert the type selector to a toggle

nvdk:
there will not be more types any time soon, so I feel this will be a bit more clear to users.


lagartoverde (approved):
If Dieter approves it for the design, for me it's good code wise looks good

PR #207, opened 2021-10-13
Added attachments number to agendapoints table

lagartoverde:
Maybe do we need some design for that number? @Dietr
If not feel free to approve and merge :)


nvdk (approved):
One thing to consider with this approach is that it can make the initial request heavy (and the cache will also be invalided more often). I guess it all depends on the amount of attachments we can expect, if it's low this will work ok.

Personally I would have gone with a separate component that fetches the attachment count for each line, but this looks fine to merge as is at the moment.

nvdk (changes requested):
missed the header, that needs to be fixed before we can merge

PR #210, opened 2021-10-13
Feature/add hash to printview

lagartoverde:


lagartoverde:
Screen Shot 2021-10-13 at 12 56 13
Looks like this, I guess we don't have to add styling as the other parts are not styled either

app/components/print-view/signature.hbs
```diff
@@ -1,4 +1,5 @@
 <div class="au-o-grid__item au-u-1-2@medium">
   <AuHeading @level="3" @skin="6" class="au-u-margin-bottom-tiny">Ondertekend door</AuHeading>
   <p>{{@fullName}} <br> {{moment-format @createdOn 'L LT'}}</p>
+  <p>Hash: {{@hashValue}}</p>
```
nvdk, line 4: I'd drop the `Hash: `

PR #213, opened 2021-11-03
Bugfix/agendapoint ordering

abeforgit:
probably fixes https://binnenland.atlassian.net/browse/GN-3037
It's hard to exactly reproduce the problem, but this pr solves 2 database inconsistency issues related to agendapoint ordering logic.
Imo the most likely cause was closing the modal while the update/delete tasks were still running. 
Due to (once again) https://github.com/machty/ember-concurrency/issues/161 this can easily cause inconsistent database state. (e.g. an agenda-item without treatment)

### DIFF
--- a/app/components/agenda-manager/agenda-context.js
+++ b/app/components/agenda-manager/agenda-context.js
@@ -102,6 +102,9 @@ export default class AgendaManagerAgendaContextComponent extends Component {
    */
   @task
   * deleteItemTask(item) {
+    // we don't use item.position here to guard against problems
+    // with position logic. The performance hit of searching here
+    // is probably minimal.
     const index = this.items.indexOf(item);
 
     this.items.splice(index, 1);
--- a/app/components/agenda-manager/agenda-context.js
+++ b/app/components/agenda-manager/agenda-context.js
@@ -74,6 +74,9 @@ export default class AgendaManagerAgendaContextComponent extends Component {
    */
   @task
   * updateItemTask(item) {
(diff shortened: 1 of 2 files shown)


app/components/agenda-manager/agenda-context.js
```diff
@@ -180,10 +181,15 @@ export default class AgendaManagerAgendaContextComponent extends Component {
       if(item.position !== index || previousItem !== previous) {
         this.setProperty(item, "position", index);
         this.setProperty(item, "vorigeAgendapunt", previous);
-        const treatment = yield item.treatment;
+        const treatment = yield item.behandeling;
```
nvdk, line 184: yikes, does this mean the order of treatments was never correctly saved?

nvdk (approved):
I'm guessing this will mitigate most issues, so this can be merged.

Couple of remarks:
- I guess it would make more sense if we could block closing the modal until our tasks are done
- if a user closes the modal before these tasks are done, will the UI look ok to them (e.g. even though the tasks haven't finished yet will they see the correct info... I'm guessing yes).

app/components/agenda-manager/agenda-context.js
nvdk: it is used here: https://github.com/lblod/notulen-prepublish-service/blob/5e95f3c3601e20b87c05b7f5e85938ce560da9ad/models/treatment.js#L75, so possible incorrect info when an extract is published

PR #222, opened 2022-01-12
Get rid of ember-cli-moment-shim and update ember-moment

abeforgit:
Adresses the v1 release of `ember-get-config` which breaks in our build. the shim package had that dependency set to any version, so a fresh install took the latest and broke our build. Luckily the package seems redundant, with the latest ember-moment working fine without it.

Draft because: more testing needed in parts where moment is used, I'm a bit strapped for time so would like someone else to look at it


lagartoverde:
I tested and it seems to work, so I will merge it on futurenow

nvdk:
I've just replaced ember-moment completely in #223, since ember-moment is a very big package and I've wanted to drop it for some time. This was a good excuse to do so

PR #223, opened 2022-01-13
replace moment with helpers based on date-fns

nvdk:
visually verified the changes and all seems to be ok, there's some small changes (we were inconsistent and sometimes formatted as `dd-MM-YYYY`, now we are consistently using `/`).

### DIFF
--- a/app/helpers/human-friendly-date.js
+++ b/app/helpers/human-friendly-date.js
@@ -5,7 +5,7 @@ import { nl } from 'date-fns/locale';
 // this is a helper to mimic the moment-calendar behaviour
 export default helper(function humanFriendlyDate([referenceDatetime]/*, hash*/) {
   //If not a date (e.g. date is undefined) return "" for printing on screen.
-  if (!(datetime instanceof Date)) return "";
+  if (!(referenceDatetime instanceof Date)) return "";
 
   try {
     return formatRelative(referenceDatetime, new Date(), { locale: nl });


package.json
```diff
@@ -126,5 +123,8 @@
     "last 1 Safari version",
     "last 1 iOS version",
     "last 1 Edge version"
-  ]
+  ],
+  "dependencies": {
+    "date-fns": "^2.28.0"
```
Windvis, line 128: Nitpick: I would move it to devDeps for consistency.

app/helpers/plain-date.js
```diff
@@ -0,0 +1,12 @@
+import { helper } from '@ember/component/helper';
+import lightFormat from 'date-fns/lightFormat';
+
+export default helper(function plainDate([datetime]/*, hash*/) {
+  try {
+    return lightFormat(datetime, 'dd/MM/yyyy');
```
Windvis, line 6: In OP we created a helper that uses [`Intl.Datetimeformat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) (with hardcoded country since we didn't have other requirements).
Maybe something similar can be done here? Or do you prefer manual formatting?

app/helpers/detailed-date.js
```diff
@@ -0,0 +1,12 @@
+import { helper } from '@ember/component/helper';
+import lightFormat from 'date-fns/lightFormat';
+
+export default helper(function detailedDate([datetime]/*, hash*/) {
+  try {
+    return lightFormat(datetime, 'dd/MM/yyyy HH:mm');
```
Windvis, line 6: I think this can be achieved with Intl.Datetimeformat as well by using some of the options:

`console.log(new Intl.DateTimeFormat('nl-BE', { dateStyle: 'short', timeStyle: 'short' }).format(date));`

app/components/app-chrome.hbs
```diff
@@ -14,7 +14,7 @@
         {{#if @editorDocument.updatedOn}}
         <li class="au-c-list-horizontal__item">
           <span class="au-c-app-chrome__status">
-            Bewaard op {{moment-calendar @editorDocument.updatedOn}}
+            Bewaard op {{human-friendly-date @editorDocument.updatedOn}}
```
Windvis, line 17: What does this look like in practice?  Doesn't this result in things like `Gisteren` or `2 dagen geleden`?

If so, doesn't `Bewaard op gisteren` sound wrong? Maybe I'm misunderstanding the output :smile:

Windvis (commented):
Looks good! I'm really curious about some built size differences :grimacing:

app/helpers/plain-date.js
nvdk: on testing I ended up with a different separator (I think `-`) and `/` was preferred here. I can recheck with customer if that's still important

app/components/app-chrome.hbs
nvdk: I agree, but this is also the case with moment-calender and for now I just want to match old behaviour

Windvis: I assumed it was a mistake in the previous implementation but fair enough :smile:

app/helpers/plain-date.js
benjay10: Not entirely: formatting as 'dd/MM/yyyy' would give '21/01/2022' and Intl gives '21/1/2022', unless you use `Intl.DateTimeFormat("nl-BE", { dateStyle: "short" }).format(datetime);`. Weird, but consistent across browsers.

app/helpers/detailed-date.js
```diff
@@ -0,0 +1,15 @@
+import { helper } from '@ember/component/helper';
+import lightFormat from 'date-fns/lightFormat';
+
+export default helper(function detailedDate([datetime]/*, hash*/) {
+  //If not a date (e.g. date is undefined) return "" for printing on screen.
+  if (!(datetime instanceof Date)) return "";
```
Windvis, line 6: I know it's merged already, but this check seems redundant since the try catch handles it already? And the comment above seems redundant as well since it's literally the transcript of what you think when you read the actual line of code :smile:.

benjay10: I thought the intention with the try...catch was to print an error in case the actual dates did not make sense, but we don't want an error when the date is just undefined, because that can happen in a normal situation. I wanted error messages only when really necessary to not cause confusion later.
Yes, the comment is pretty useless like this, sorry for that, but at least it means this check was put there for a reason.

PR #225, opened 2022-01-18
add support for mobility measures (IRGN)

nvdk:


Dockerfile
```diff
@@ -3,10 +3,10 @@ FROM madnificent/ember:3.20.0 as builder
 LABEL maintainer="info@redpencil.io"
 
 WORKDIR /app
-COPY package.json .
-RUN npm install
+COPY package.json package-lock.json ./
+RUN npm ci
 COPY . .
-RUN ember build -prod
+RUN ember build
```
nvdk, line 9: we should try to get this fixed still, I believe this was due to an issue with the datastore not working correctly in the editor?

app/components/rdfa-editor-container.js
```diff
@@ -4,7 +4,12 @@ import { tracked } from '@glimmer/tracking';
 
 export default class RdfaEditorContainerComponent extends Component {
   @tracked editor;
-
+  plugins = [
+    'roadsign-regulation',
+    'template-variable',
+    'besluit',
```
nvdk, line 10: might want to tweak the order here a bit

nvdk:
todo:
 - [ ] inserting a template (e.g. when you create a new agendapoint from the meeting) doesn't close the menu
 - [ ] the decision type plugin doesn't collapse correctly, it should collapse when a type is set (if a subtype is required, only when the subtype is set)

PR #243, opened 2022-02-16
Linting after upgrade 3.28

benjay10:
After the upgrade to Ember v3.28, Eslint/Prettier was re-enabled. This PR is the result of beautifying the code with these tools on their default mostly settings according to ember new.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,24 +10,18 @@ module.exports = {
       legacyDecorators: true,
     },
   },
-  "globals": {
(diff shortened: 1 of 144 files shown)


benjay10:
Note that this PR is based on the previous Ember upgrade branch.

benjay10:
There are still linting issues: some mixins from ember-data-table are still in use and cause warnings on Eslint (see config).

nvdk (approved):

PR #244, opened 2022-02-17
Fixing production builds "out of memory" problems

benjay10:
The option `collapse_vars` was disabled on Terser and this caused the RAM usage to spike for some reason, failing the build due to out of memory problems. Removing this option leaves Terser with its own defaults and that seems to produce correct builds.


nvdk:
Thanks for investigating this! Probably best to also remove the node options from the Dockerfile if we no longer need it.

PR #245, opened 2022-02-25
ember resources experiment

Asergey91:
So I have experimented with ember resources and rewrote 2 components one of them relies on tracked properties that are passed down (attachment-number) ther other one just needs to fetch stuff (administrative-body-select).I tried both tasks and async functions.

So this cuts down on code and does solve a couple of our issues in how we deal with async requests in components.

However there is definately some wierdness that I ran into even with these basic use cases. The most obvious one is having to add `await Promise.resolve();` for seemingly no reason at the top of the async function (... ok there is a reason and you can read here https://github.com/ember-learn/guides-source/issues/1769) .

But overall I think this is an improvement

### DIFF
--- a/app/components/administrative-body-select.hbs
+++ b/app/components/administrative-body-select.hbs
@@ -2,7 +2,7 @@
   <PowerSelect
     @placeholder={{t "manageZittingsData.selectPlaceholder"}}
     @selected={{@selected}}
-    @options={{this.administrativeBodyOptions}}
+    @options={{this.administrativeBodyOptions.value}}
     @onChange={{@onChange}}
     @triggerId={{@id}}
     as |administrativeBody|
--- a/app/components/administrative-body-select.js
+++ b/app/components/administrative-body-select.js
@@ -1,9 +1,8 @@
 import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
-import { task } from 'ember-concurrency';
 import sub from 'date-fns/sub';
 import isAfter from 'date-fns/isAfter';
+import { trackedFunction } from 'ember-resources';
 
 const VALID_ADMINISTRATIVE_BODY_CLASSIFICATIONS = [
   'http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5ab0e9b8a3b2ca7c5e000005', //	"Gemeenteraad"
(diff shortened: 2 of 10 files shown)


abeforgit (commented):
@benjay10 you mentioned in your commit message that you think the resources in the meeting form are messy, and wonder if there's a cleaner way

I think `ember-data-resources` is probably what we want. Afaict, it gets around the awkwardness of the `Promise.resolve()` workaround with respect to ember-data's methods. 
In other situations, it's on us to be aware of what it means if we assign a value to or read from a tracked property. This is not new, we have to do this in getters already. You cannot write to a tracked value _after_ you read from it in a getter, or it causes an infinite loop ( https://www.pzuraq.com/blog/how-autotracking-works#principle-1-predictable-output ). This is the same underlying issue as is happening with resources, although it manifests differently because of asynchronicity (the yield closes the tracking frame, so it _appears_ you can write-after-read, even though you're not)

About the messiness: have a little monologue :stuck_out_tongue:, free of charge:

First of all, I see that you could consider the previous version cleaner (though I would disagree), _if you ignore it was wrong_
The contract that this component implies is this: you give it a meeting, and it renders it. The end. That rendering is complicated, and involves a bunch of async stuff, but really, thats what it does.

Now, crucially, the whole point of modern frontend frameworks is to manage _when_ this is done. Sure, on loading the page. But _also_ whenever its arguments change. This component did not do that before. Sure, it _worked_. But that's because we only use this component in 1 place, namely a route template. It just so happens that whatever this component depends on was not a value that was changed dynamically, it was always derived from the URL.

I think it's worth reflecting here, _why_ should this component be reactive? we use this component simply to organize template and loading logic, since otherwise our route and controller would become unwieldy. State that is derived _only_ from the route URL can be considered static. Or rather, that state is _automatically_ reactive, just because thats how browsers work. So why bother thinking about reactivity in a component that will never rerender in the context of a single pageload?

In my opinion, the answer is obvious: for _our_ sanity. Rather than trying to remember if the arguments you are working with are coming down from the route, and can thus be considered static, or if they can possibly change through some user interaction or other trigger, is madness.
Make _everything_ reactive. In a lot of cases, you don't need resources for this. Getters can get you surprisingly far, and the ability of the rendering engine to understand promises gets you the rest of the way to 90% of usecases.

However, we happen to often need that other 10%. And in that situation, Ember makes it _hard_ to make your components truly reactive, because it is missing a chunk of tools to do it. It has _most_ of the tools, but not all of them. This is why `render-modifiers` is such a ubiquitous addon.

With resources, we can finally uphold the implicit contract of reactivity in each component,  without awkwardness in the templates. You can now consider the meetingform in isolation. You _know_ the rendered values will always be correct, _regardless_ of where or how this component is used. You can think of the component as a function of arguments to html again.

Oh, and if the arguments never happen to change? No harm done. The logic will only execute once anyway!

PR #246, opened 2022-03-03
bump template and decision type plugin

nvdk:
bumps editor to 0.50.0 and moves the template and decision plugin to the latest versions. updated to the new plugin wiring in GN.

edit: as mentioned in the chat  I think it's odd for users that the "voeg sjabloon toe" button dissapears while typing, might be better to disable it


nvdk:
Hmm yes that's the old way of calling the plugin (the legt one). This is fixed in https://github.com/lblod/ember-rdfa-editor/pull/232 . perhaps it's best to convert this to a draft until that's merged

abeforgit (approved):
Bit weird that this pr also includes a release? but looks fine otherwise

nvdk:
> Bit weird that this pr also includes a release? but looks fine otherwise

hmm yes, probably because I branched of master and we don't have the habit of merging the releases back into dev

benjay10 (approved):

nvdk:
rebased on latest dev

PR #247, opened 2022-03-03
make the template required when creating an agendapoint

nvdk:


lagartoverde (changes requested):
Screen Shot 2022-03-03 at 15 25 05
Code looks good and it works, but I think we should add the required tag in the UI when creating the agendapoint

benjay10 (approved):
Seems to work correctly!

PR #261, opened 2022-03-30
ran npm update

nvdk:
+ release-it@14.13.1
+ eslint-config-prettier@8.5.0
+ ember-modifier@3.2.0
+ qunit@2.18.1
+ ember-auto-import@2.4.1
+ ember-power-select@5.0.4
+ ember-simple-auth@4.2.1
+ prettier@2.6.1
+ sass@1.49.9
+ webpack@5.70.0
+ @lblod/ember-acmidm-login@1.4.0
+ ember-unique-id-helper-polyfill@1.1.0


nvdk:
> I think this is not a good strategy. Doesn't using "^" already get you the latest version of a package in the range of the major version when using `npm install`? So "^2.5.1" will already get version "2.6.1" for you. In that case, none of the changes in this commit will do anything useful. We should instead update the `package-lock.json` file and use `npm ci` everywhere else to make sure the packages that we test with are the ones used in production, so we know they work.

this is a misunderstanding on your part, npm update explicitly updates in both package.json and package-lock.json

nvdk:
> Furthermore, some of these packages are identical to the defaults of [ember-new-output](https://github.com/ember-cli/ember-new-output/blob/v3.28.5/package.json) and I would prefer to keep them that way. We know those version are tested for that specific Ember version and are nicely compatible. We'll update those when moving to newer Ember versions.

this I can agree with

benjay10:
> this is a misunderstanding on your part, npm update explicitly updates in both package.json and package-lock.json

Not a misunderstanding: I know that is what it does. I wanted to suggest to do `npm install` regularly to keep the `package-lock.json` up to date with the latest packages within semver range instead of using `npm update`. I prefer not to change the `package.json` unless there is a specific reason to (e.g. we really need a new minimum version of the editor).

nvdk:
> > this is a misunderstanding on your part, npm update explicitly updates in both package.json and package-lock.json
> 
> Not a misunderstanding: I know that is what it does. I wanted to suggest to do `npm install` regularly to keep the `package-lock.json` up to date with the latest packages within semver range instead of using `npm update`. I prefer not to change the `package.json` unless there is a specific reason to (e.g. we really need a new minimum version of the editor).

as explained in my edit `npm install` doesn't update your package-lock.json file

benjay10:
> as explained in my edit `npm install` doesn't update your package-lock.json file

So it does, but only if there is a mismatch of versions.

I still think there could be alternative solutions, like rebuilding the full `package-lock.json`, so that we don't have to cherrypick newer versions in the `package.json`.

nvdk:
> > as explained in my edit `npm install` doesn't update your package-lock.json file
> 
> So it does, but only if there is a mismatch of versions.
> 
> I still think there could be alternative solutions, like rebuilding the full `package-lock.json`, so that we don't have to cherrypick newer versions in the `package.json`.

we could definitely go our own custom way, but there seems to be little benefit in that.

PR #263, opened 2022-04-01
improve import flow using updated import plugin

nvdk:
this depends on https://github.com/lblod/ember-rdfa-editor-import-snippet-plugin/pull/2


lagartoverde (approved):
Works correctly and code looks good, good job 👍

PR #264, opened 2022-04-01
Fixed off by one errors on adding agendapoints

benjay10:
We can now add agendapoints to the beginning or end of the list when there is at least 1 agendapoint (previously 2), and the agendapoints now show correctly at the end of the list when chosen so. These were "off by one" errors on collections.


lagartoverde (approved):
Weird that we have missed this until now, it works good and looks good

PR #268, opened 2022-04-25
Fix alignment of datepicker dialogs in modal windows

Dietr:
Fixes the alignment of the datepicker dialog in:
- new meeting modal
- meeting > general information modal

### DIFF
--- a/app/components/attachments-number.js
+++ b/app/components/attachments-number.js
@@ -16,6 +16,6 @@ export default class AttachmentsNumberComponent extends Component {
   @restartableTask
   *getAttachmentsNumber() {
     const attachments = yield this.args.documentContainer.attachments;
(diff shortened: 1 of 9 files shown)


benjay10 (changes requested):
Looks like a good idea, and indeed works nicely on larger monitor, but when testing on a smaller screen (smaller and taller screens, like smartphones), the datepicker flies out of the modal's view and can not be interacted with.
See attached screen recording (if that works).

https://user-images.githubusercontent.com/12394202/166339262-d2cacb91-7baf-4163-900a-d3bc4f7cf5f0.mp4

Dietr:
Added a fix for smaller screens. Apparently the default positioning of the datepicker was also kind of broken.

benjay10 (approved):
Nice! This looks very clean

PR #269, opened 2022-04-27
fix attachments number in the agendapoints screen

lagartoverde:
Fixes https://binnenland.atlassian.net/browse/GN-3396, basically the meta count is not updated but the length of the attachments is.
Maybe also something we should check with Aad


lagartoverde:
I did it like this, as the other attachments count are using the length

nvdk (approved):
I don't think relationships have meta info at all :+1:

PR #271, opened 2022-04-27
Removed style that displayed article number as block

lagartoverde:
Fixes https://binnenland.atlassian.net/browse/GN-3417.
I didn't find any use for that block of css but you will know better than me


Dietr (approved):

nvdk:
@lagartoverde We will also need to fix this in GN publication frontend

PR #273, opened 2022-05-04
changing the geplandOpenbaar of an agendapoint affects the openbaar of a behandeling

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-3425

### DIFF
--- a/app/components/agenda-manager/edit.js
+++ b/app/components/agenda-manager/edit.js
@@ -28,11 +28,6 @@ export default class AgendaManagerEditComponent extends Component {
   }
 
   @action
-  updateGeplandOpenbaar() {
-    this.oldGeplandOpenbaar = this.args.itemToEdit.geplandOpenbaar;
-  }
-
-  @action
   cancel() {
     this.args.onCancel();
   }
--- a/app/components/agenda-manager/edit.js
+++ b/app/components/agenda-manager/edit.js
@@ -5,15 +5,33 @@ import { inject as service } from '@ember/service';
 
 export default class AgendaManagerEditComponent extends Component {
   @service documentService;
-
+  oldGeplandOpenbaar;
+  constructor() {
+    super(...arguments);
+    this.oldGeplandOpenbaar = this.args.itemToEdit.geplandOpenbaar;
+  }
   get isNew() {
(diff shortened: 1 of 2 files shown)


benjay10 (commented):
I don't understand what the `did-update` modifier is trying to achieve. I could not trigger it. Can someone explain?

Also, can the modifier be put on the top level component in this file? The effect should be the same, but that would make this PR much cleaner, by just adding 1 single line in this file.

nvdk:
> I don't understand what the `did-update` modifier is trying to achieve. I could not trigger it. Can someone explain?
> 
> Also, can the modifier be put on the top level component in this file? The effect should be the same, but that would make this PR much cleaner, by just adding 1 single line in this file.

I'm not sure if it's needed here. The did-update notifier is typically useful if you have an argument passed into a component and want to trigger some function if the the value of the argument is updated. it can't be added to a component because it will be considered an attribute that's passed down into the component. However it could be added on another element, I don't think a wrapping div is required here and I also dislike the addition.

@lagartoverde Can you elaborate why the did-update modifier is needed here?

app/components/agenda-manager/edit.js
```diff
@@ -5,15 +5,33 @@ import { inject as service } from '@ember/service';
 
 export default class AgendaManagerEditComponent extends Component {
   @service documentService;
-
+  oldGeplandOpenbaar;
```
nvdk, line 8: I think this can just be a getter and then the did-update can be dropped altogether

app/components/agenda-manager/edit.js
```diff
@@ -5,15 +5,33 @@ import { inject as service } from '@ember/service';
...
+  constructor() {
+    super(...arguments);
+    this.oldGeplandOpenbaar = this.args.itemToEdit.geplandOpenbaar;
```
abeforgit, line 11: why are you not just using a getter?

abeforgit (changes requested):

lagartoverde:
I removed the did-update modifier as you didn't like it

app/components/agenda-manager/edit.js
lagartoverde: Because I want to get the original value when the component was created not the actual value of the variable

PR #276, opened 2022-05-16
added validation to meeting end date/time

Asergey91:
solves GN-3418


nvdk (changes requested):
Fails on linting rules, I recommend setting up prettier in your editor or adding the lint check to your git pre-commit hook

lagartoverde (approved):
Looks good and the listing error is gone 👍

PR #280, opened 2022-06-06
improve error reporting if preview of notulen fails

nvdk:


abeforgit (approved):
Code looks good and checks out with ember-conc docs, but havent tested

PR #285, opened 2022-06-09
Introducing ember-plausible

benjay10:
This PR removes old analytics set-up based on Plausible directly and introduced the ember-plausible addon. Analytics are enabled on all routes by default.


nvdk (commented):
in the interest of backwards compatibility (and not having to change the env vars on all servers) I would keep using `EMBER_ANALYTICS_APP_DOMAIN`. If we then also provide a default for conf.apiHost in the application route no config changes need to happen on the server.

nvdk (approved):

benjay10:
@nvdk Do you want to include some default URL for the analytics endpoint in the code/config?

PR #286, opened 2022-06-09
Plausible custom event 'Create agendapoint'

benjay10:
This PR includes the changes to post analytics about a custom event for creating an agendapoint and includes the template id and the title. I added the selected template as an argument to the onCreate action so that I could write the analytics code in the controller instead of the component.

**The custom events could not be checked on the dashboard yet.** I did not have access to the dashboard's configuration to inspect if the custom event's properties have correctly been logged. The public dashboard only shows the goal, so I know it works at least.


benjay10:
Also, this PR is based on another PR where ember-plausible is introduced. See: https://github.com/lblod/frontend-gelinkt-notuleren/pull/285

nvdk:
image
can confirm it works at least. FYI, you can see the properties by clicking on the goal, I think that should also work in the shared URL.

nvdk (approved):

nvdk:
I get the reflex to include the id, but I don't think it has much value in the analytics tools to be honest

benjay10:
@nvdk I thought I tried clicking on the Goal, but no :facepalm: It indeed shows the properties.

Including the id is for easier data processing later, but it shoud be one or the other, not both. I'll go ahead and remove the ID then.

PR #289, opened 2022-06-15
update package version of citaten-plugin

elpoelma:
This update fixes an issue with willDestroy-hooks not being triggered as the previous version of this package caused an exception.


abeforgit (approved):
You can just merge this I think

PR #290, opened 2022-06-17
bump ay11-refocus to 3.0.0

nvdk:


Dietr:
I don't think we use this package. Probably some legacy as it is now handled by this component: https://github.com/appuniversum/ember-appuniversum/blob/development/addon/components/au-navigation-narrator.hbs

nvdk:
aha even better, so we could drop it entirely? I think we have to replace the component [here](https://github.com/lblod/frontend-gelinkt-notuleren/blob/b3058869d0d49bf45f0a079f2a87f5540878528e/app/components/vo-page.hbs#L8) with the one from AU then?

Dietr:
Missed that one. Now I remember why it's used. We should keep it, still not the best solution for these informative pages (login, disclaimer, etc.) but I see no other way to reset the focus to the top of the page.

PR #296, opened 2022-06-30
Bugfix/search and pagination

Asergey91:


nvdk (changes requested):
looks good, but you can simplify the updateFilter by using `@restartableTask` and timeout helper of ember-concurrency, so you no longer need to keep track of `this.debouncing` and the timeout helper. See http://ember-concurrency.com/docs/examples/autocomplete for an example.

nvdk (approved):

PR #297, opened 2022-07-05
Create revision layout

Dietr:
Rework the revision page to fit into the design.

## Needs dev work/help
- [x] "Bijlagen" button is not clickable from Versions tab
- [x] "Exporteer als html" in versions tab is a mockup > needs the actual functionality
- [x] Make the versions tab also available in the meeting agendapoint modal


elpoelma:
Works well overall, maybe the 'Bestand acties' dropdown could contain entries in order to export the html of the current version or remove a specific version. Additionally, it would be nice that the 'Herstel versie' button is disabled when the most recent version is selected.

nvdk:
the custom edit view in meeting is now gone, I've discussed this with dieter. 

two things I've noticed: 
- [ ] attachment / revision count is not updating after uploading / saving.
- [ ] getting two warnings instead of one when leaving agendapoint without saving

app/controllers/agendapoints/edit.js
```diff
@@ -89,6 +85,7 @@ export default class AgendapointsEditController extends Controller {
       editorDocument.updatedOn = new Date();
       editorDocument.title = this.editorDocument.title;
       editorDocument.previousVersion = this.editorDocument;
+      editorDocument.documentContainer = this.documentContainer;
```
nvdk, line 88: this might get lost in the big lump, but this is a bug that will need a migration to correct on all environments. There's supposed to be a direct link between any version and the container and that was no longer being set.

elpoelma:
> the custom edit view in meeting is now gone, I've discussed this with dieter.
> 
> two things I've noticed:
> 
>     * [ ]  attachment / revision count is not updating after uploading / saving.
> 
>     * [ ]  getting two warnings instead of one when leaving agendapoint without saving

I created a solution in order to prevent the double warnings on #342 .

PR #298, opened 2022-07-05
Changes for new besluit-type-plugin

benjay10:
This PR removes the old TTL files with the data about the BesluitTypes. The data will be fetched from a SPARQL endpoint in the future. The new besluit-type-plugin ([not released yet](https://github.com/lblod/ember-rdfa-editor-besluit-type-plugin/pull/23)) will do this.
It also has to include some Webpack configuration copied from ember-rdfa-editor-besluit-type-plugin to package module `process`. All tested via `npm link` and it seems to work.

**NOTE:**
* The SPARQL endpoint is to be configured via variables. These are used and defined in the plugin, but do show up correctly in the `index.html`.


benjay10:
No idea why the Drone integration testing failed before (it complained about Missing npm packages, while it is supposed to do an `npm ci`). At least that is fixed now with the inclusion of the new besluit-type-plugin that has been released.

PR #315, opened 2022-08-19
Bugfix/make meeting date more clear

nvdk:
since our users are confused by the dates (often only changing the planned date), this should make things a bit more clear.


lagartoverde:
Screen Shot 2022-08-19 at 14 30 55
Getting this error when trying to change the end date

nvdk:
good catch @lagartoverde, should be fixed now

PR #316, opened 2022-08-19
Bugfix/make meeting date more clear

nvdk:


nvdk:
Copying hotfix from #315 to mainline

PR #319, opened 2022-09-01
Print preview improvements

Dietr:
Improvements on native printing on all screens:
- consistent font sizing
- reduce overall white space for reduced size of output


abeforgit (approved):
Looks much better, although it's impossible to tell whether this will satisfy the user

PR #320, opened 2022-09-06
Ensure meeting is deleted from cache before returning to meeting overview when deleting a meeting

elpoelma:
This PR adds polling in order to check that a deleted meeting has been removed from the cache before returning to the meeting inbox. Solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3569. Implemented with fetch instead of ember-data.


app/components/delete-meeting.js
```diff
@@ -15,9 +15,23 @@ export default class DeleteMeetingComponent extends Component {
...
   @task
   *deleteMeeting() {
+    const meetingId = this.args.meeting.id;
     yield this.args.meeting.destroyRecord();
     this.displayDeleteModal = false;
-    this.router.transitionTo('inbox.meetings');
+    setTimeout(() => {
+      const pollingInterval = setInterval(async () => {
+        try {
+          const response = await fetch(`/zittingen/${meetingId}`);
```
abeforgit, line 24: So `findRecord` fails because the store already knows the record is deleted and refuses to query again.
But could you try `this.store.query` here? It's more in line with the rest of the app and might make maintenance easier down the line (since it'll be model-aware)

abeforgit (changes requested):

abeforgit:
Overall logic is sound tho, so if my suggestion doesn't work (or if you don't agree with it) you can merge as-is

app/components/delete-meeting.js
elpoelma: ```this.store.query``` does not seem to work as I cannot correctly pass the meeting id to this method. Same issue with the ```queryRecord``` method.

elpoelma: Eventually I settled on using `fetch` as it was faster. Additionally `this.store.query` contained some issues while resolving it when the server returned a 500 status code.

PR #323, opened 2022-09-07
Add release task to package.json

abeforgit:
I think having the release command as part of the project scripts makes way more sense


abeforgit:
TODO: if we agree on this, we should update the release guide in gitbook

elpoelma (approved):

PR #325, opened 2022-09-12
Feature/regulatory statements

nvdk:
reworking #317 a bit for a cleaner diff, needs https://github.com/lblod/app-gelinkt-notuleren/pull/111 as a backend


app/components/document-creator.js
```diff
@@ -81,7 +81,7 @@ export default class DocumentCreatorComponent extends Component {
     if (this.template) {
       if (this.template.reload) {
         // regular templates from templatesForContext do not return body of template
-        await this.template.reload();
+        await this.template.reload(this.template);
```
nvdk, line 84: this is odd, reload doesn't require a param afaik:
https://api.emberjs.com/ember-data/3.28/classes/Model/methods/reload?anchor=reload

app/services/regulatory-attachments-fetcher.js
```diff
@@ -51,8 +53,14 @@ export default class RegulatoryAttachmentsFetcher extends Service {
       const templates = bindings.map((binding) => ({
         container: binding.container.value,
         title: binding.title.value,
+        reload: async (template) => {
+          const response = await fetch(
+            `${config.regulatoryStatementPreviewEndpoint}/${binding.uuid.value}`
```
nvdk, line 58: as discussed I think this should use the regular file service

abeforgit:
how does this relate to the futurenow branch?

nvdk:
they should be equal, any dev work should happen on this branch. futurenow is just a deploy branch as far as I'm concerned

abeforgit:
this can now be tested with the dev backend

nvdk:
We probably want to implement https://binnenland.atlassian.net/browse/GN-3691 before actually merging this, but that could also be a separate PR.

lagartoverde (approved):
Works as expected, but the table of contents plugin is not updating at all

elpoelma:
> Works as expected, but the table of contents plugin is not updating at all

I will look into it.

elpoelma:
The issue with the table of contents not updating with the template content lies in the article-structure-plugin. The main issue is that two different divs with the type say:DocumentContent are created:
- one div with the structure of the template
- one div with the 'Insert here' text
The problem is that the first div with say:DocumentContent does not contain the necessary prefixes for the editor to resolve the rdfa of its content, that is why the table of contents can not detect it.

elpoelma:
This PR for the registry: https://github.com/lblod/frontend-reglementaire-bijlage/pull/27 adds new prefixes to the root of the document so the rdfa content is correctly resolved. It also changes the type-uri for a reglementaire bijlage.

abeforgit (changes requested):
Just plain doesn't work, not sure why this was approved
creating a  template errors out
and this is with setting the endpoints correctly to the dev server

elpoelma:
> doesnt work (anymore) creating a template errors out and this is with setting the endpoints correctly to the dev server

Locally it seems to work for me, what specific errors do you get? JSON parsing errors?

abeforgit:
yep my bad, set up the wrong preview endpoint

app/services/regulatory-attachments-fetcher.js
abeforgit: created https://binnenland.atlassian.net/browse/GN-3732

app/components/document-creator.js
abeforgit: aah but here's the fun part
template is not an ember-data model here
look at the RegulatoryAttachmentsFetcher service
it makes a raw sparql request and from the bindings it creates an object which _does_ have a reload method that takes an argument

lagartoverde: Ah yes sorry, forgot about this comment, basically I wanted to make it compatible with the old document creator and it was the best way I came up with

lagartoverde: If anyone has a better idea feel free to suggest it

nvdk: but, why? we already check if the object has a reload function and only execute reload then?

lagartoverde: Because I need a hook to fetch the template content, so I don't fetch all the contents before the user selects the one he wants to use.

app/routes/inbox/regulatory-statements.js
```diff
@@ -0,0 +1,28 @@
...
+
+  queryParams = {
+    page: { refreshModel: true },
+    sort: { refreshModel: true },
+    filter: { refreshModel: true },
+  };
+
+  async model(params) {
+    const options = {
+      sort: params.sort,
```
abeforgit, line 16: this view should sort on date created by default
https://binnenland.atlassian.net/browse/GN-3735 tracked here if too much work

abeforgit (changes requested):

abeforgit:
about the endpoints, do we add the dev backends in the dev env like we do for roadsigns?
I could see it being annoying if you want to override that backend with an env var on your local, but on the other hand, how often would that happen anyway

app/routes/inbox/regulatory-statements.js
abeforgit: https://github.com/lblod/frontend-gelinkt-notuleren/pull/336

PR #330, opened 2022-09-29
Fix selection errors when trying to save with empty selection

abeforgit:
description in progress


elpoelma (approved):
Agendapoint saves correctly without any selection errors.

nvdk:
I'm merging this and deploying to QA so we can include this in the production release

PR #331, opened 2022-09-30
Fixed bug where the attachment wasn't assigned to the first decision when inserted

lagartoverde:
What it looked like a new feature ended up being a bugfix, we already had the code for it, but it wasn't working as expected


nvdk (approved):
works!

PR #333, opened 2022-10-04
Filter based on validThrough property

lagartoverde:
First of all this depends on https://github.com/lblod/reglement-publish-service/pull/5 so we should wait for that to be merged and deployed, but I created the PR so we can start a little discussion.
I did the most basic thing, if the reglement has an expiration date not show it, I was thinking on filtering based on the ones that are still valid (even thought we don't really set this validthrough date in the future, we just set it to today when we delete in the regulatory statements app). This would be more semantically correct, but also would bring the question: "Do we want to allow the user to include a template that will expire tomorrow, in the next week...?", I think that for the moment is better to leave it like this and improve the flow in the future, but it's open for discussion

### DIFF
--- a/app/services/regulatory-attachments-fetcher.js
+++ b/app/services/regulatory-attachments-fetcher.js
@@ -28,9 +28,10 @@ export default class RegulatoryAttachmentsFetcher extends Service {
           ext:currentVersion ?container.
         ?container dct:title ?title.
         ?reglement ext:publishedVersion ?publishedContainer.
-        FILTER NOT EXISTS {
+        OPTIONAL { 
           ?reglement schema:validThrough ?validThrough.
         }
+        FILTER( ! BOUND(?validThrough) || ?validThrough < NOW()) 
       }
     `;
     const details = {
--- a/app/services/regulatory-attachments-fetcher.js
+++ b/app/services/regulatory-attachments-fetcher.js
@@ -31,7 +31,7 @@ export default class RegulatoryAttachmentsFetcher extends Service {
         OPTIONAL { 
           ?reglement schema:validThrough ?validThrough.
         }
-        FILTER( ! BOUND(?validThrough) || ?validThrough < NOW()) 
+        FILTER( ! BOUND(?validThrough) || ?validThrough > NOW()) 
       }
     `;
     const details = {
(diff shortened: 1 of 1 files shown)


abeforgit:
I would say no matter what we do, I would compare the dates. Like you said that should have the same effect since we're only setting current date when deleting, but just imagine we come back to this in 2 months time and see a property called "validThrough", we'd expect it to work as the name implies.

abeforgit:
the pr you mentioned is now merged and deployed on dev.reglementairebijlagen.lblod.info

abeforgit:
Well it just kinda doesn't work, deleted templates still show up

abeforgit (changes requested):
Well it just kinda doesn't work, deleted templates still show u

app/services/regulatory-attachments-fetcher.js
```diff
@@ -21,11 +21,17 @@ export default class RegulatoryAttachmentsFetcher extends Service {
...
       select distinct * where {
         ?publishedContainer a ext:PublishedRegulatoryAttachmentContainer;
           mu:uuid ?uuid;
           ext:currentVersion ?container.
         ?container dct:title ?title.
+        ?reglement ext:publishedVersion ?publishedContainer.
+        OPTIONAL { 
+          ?reglement schema:validThrough ?validThrough.
+        }
+        FILTER( ! BOUND(?validThrough) || ?validThrough < NOW())
```
nvdk, line 34:
```suggestion
        FILTER( ! BOUND(?validThrough) || ?validThrough > NOW()) 
```
I guess valid through should be greater than the current time for the template to be valid

nvdk (changes requested):

app/services/regulatory-attachments-fetcher.js
lagartoverde: I received the email just as I was pushing the commit :P

lagartoverde:
> Well it just kinda doesn't work, deleted templates still show up

It should work now I fixed a typo in the query, also there's a looong waiting time for the cache

abeforgit:
> > Well it just kinda doesn't work, deleted templates still show up
> 
> It should work now I fixed a typo in the query, also there's a looong waiting time for the cache

you can just go to the server and restart the sparql-cache service, that's what I've been doing

abeforgit (approved):
works now

PR #334, opened 2022-10-04
Update toc plugin which fixes issues with overflowing entries

elpoelma:
Solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3656.


abeforgit (approved):
works afaict

PR #337, opened 2022-10-06
Save document title after edit

elpoelma:
This ensures that the title of the document is saved just after having edited it. An additional save of the document is no longer required.


nvdk:
one consideration is that this doesn't create a new version of the document, which is different from before. open question whether this is what a user expects. (phrased another way: do title changes warrant a new version?)

nvdk:
@elpoelma can you fix the merge conflict?

elpoelma:
> @elpoelma can you fix the merge conflict?

Done, can this be merged?

nvdk:
I'm hesitant because of the above remark :)

abeforgit:
semantically, I'd say yes, title changes warrant a new version
versions are pretty transparent to the user anyway

nvdk (changes requested):
looked into this a bit more and we were already saving on onSubmit as far as I can see, so we should clean this up. because now we are calling the component with 
```js
      <EditorDocumentTitle @title={{@editorDocument.title}} @editActive={{@editorDocument.isNew}} 
        @onChange={{this.updateDocumentTitle}} @onSubmit={{this.saveDocument}}  @onCancel={{this.resetDocument}} 
```
and both onSubmit and onChange save the document ? at the very least this should be cleaned up.

nvdk:
I've reworked this a bit + fixed some other stuff I ran into, needs a proper review @elpoelma :)

app/components/app-chrome.js
```diff
@@ -19,7 +19,7 @@ export default class AppChromeComponent extends Component {
   }
 
   @action
-  updateDocumentTitle(title) {
-    this.args.editorDocument.title = title;
+  async resetDocument() {
```
nvdk, line 22: woops I think this bit is no longer relevant and can go

elpoelma:
There was some small issue in the `cancel` method, otherwise now it seems to work well. TItle edits also add a new version correctly now which shows up in the revisions tab.

elpoelma:
I'll update the reglementaire bijlage frontend in order to have the same workflow tomorrow.

app/components/editor-document-title.js
```diff
@@ -5,27 +5,43 @@ import { tracked } from '@glimmer/tracking';
...
 
     if (title) {
       this.error = false;
     }
   }
 
+  @action
+  submit(event) {
+    event.preventDefault();
+    console.log(event, this.title);
```
nvdk, line 31:
```suggestion
```

PR #341, opened 2022-10-12
Fix regex which matches for generateUuid() in the template

elpoelma:
The previous regex did not escape some characters which caused the match to fail. Partly a solution for https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3682.


elpoelma:
Additionally, this PR also solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3765

nvdk (approved):
works!

PR #342, opened 2022-10-13
Fix: prevent the confirm dialog to show up twice when exiting without saving

elpoelma:
This PR checks if a transition target is a loading route when trying to show a confirm dialog. When it is a loading route, the dialog isn't shown.


nvdk:
interesting that the component isn't destroyed yet while the next route is already in a loading state. Great find! thank you :)

PR #344, opened 2022-10-14
fix deprecations

usrtim:


nvdk (approved):
looks good to me, thanks for this!

abeforgit (changes requested):
Looks good, but I spotted a few more:

missed deprecations in the following pages:
- inbox/trash route: implicit store injection
- meetings/<id>/publish/uittreksels: implicit store injection
- all publication pages: alertIcon, alertTitle and alertSkin. To be investigated

Remaining deprecations that need further investigation, but are probably not our fault:
- ember-modifier related deprecations on almost every page
- deprecation in "draggable-object.hbs", probably coming from `ember-drag-drop`

Some remaining deprecations come from the editor: 
- Appuniversum "AuToolbar" and "AuList" deprecations on pages with an editor

abeforgit (approved):
Looks great now!

PR #345, opened 2022-10-14
Fix infinite pages when printing a meeting

Dietr:
Reset table styles on print: seems to be the cause of the infinite pages issue on print.

Needs thorough testing as the issue is not always present. In my tests it always showed when there are votes made with an agendapoint.


nvdk:
FYI I've created a hotfix build out of this and deployed it to gebruikerssessie for testing

PR #348, opened 2022-10-19
Use file service from regulatory-statements application in order to fetch template

elpoelma:
This PR includes changes on how the regulatory statement templates are fetched:
- The SPARQL query used to fetch a template has been modified in order to conform to the new data model.
- The template content is now fetched from the regulatory statements file service instead of the custom preview endpoint.
- The `regulatoryStatementPreviewEndpoint` environment variable has been removed and replaced in favor of `regulatoryStatementFileEndpoint`. Both regulatory-statement endpoint ENV vars have been set to fetch from localhost in order to test this PR.

This PR relies on https://github.com/lblod/app-reglementaire-bijlage/pull/27.
This PR is a solution to https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3732.

TODO:
- [ ] The SPARQL query should be updated with the correct namespaces (instead of `ext`).

### DIFF
--- a/app/services/regulatory-attachments-fetcher.js
+++ b/app/services/regulatory-attachments-fetcher.js
@@ -24,12 +24,11 @@ export default class RegulatoryAttachmentsFetcher extends Service {
       PREFIX schema: <http://schema.org/>
       select distinct * where {
         ?publishedContainer a ext:PublishedRegulatoryAttachmentContainer;
-          mu:uuid ?uuid;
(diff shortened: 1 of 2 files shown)


lagartoverde (approved):
Looks good, remember to change the config when deploying

lagartoverde:
We still have to think about the reload problem, but I think that's not the point of this PR

PR #349, opened 2022-10-19
Add export-to-html menu action to regulatory-statements document view

elpoelma:
Solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3771.


abeforgit:
Some notes, which may or may not be an issue:
- table of contents does not get exported
- we lose the `lblodgn:ReglementaireBijlagen` type, since that type is defined outside the editor root. this is probably fine, and will improve when we rework the way we do types in documents

abeforgit (approved):

abeforgit:
imo you can merge without it

PR #353, opened 2022-11-02
Fix issue with voting overview page where the number of votings are limited to 20

elpoelma:
This PR includes two fixes:
* For a specific treatment of an agendapoint, all votings are fetched and not only the first 20.
* The positions of the votings are fixed automatically when a voting is removed in the middle of the list.

This PR is to be used in combination with: https://github.com/lblod/app-gelinkt-notuleren/pull/118 and solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3798

### DIFF
--- a/app/components/treatment/voting/modal.js
+++ b/app/components/treatment/voting/modal.js
@@ -41,8 +41,8 @@ export default class TreatmentVotingModalComponent extends Component {
       sort: 'position',
       page: {
         size: pageSize,
-      }
-    })
+      },
+    });
     const count = firstPage.meta.count;
     firstPage.forEach((result) => stemmingen.push(result));
     let pageNumber = 1;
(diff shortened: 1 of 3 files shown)


app/components/treatment/voting/modal.js
```diff
@@ -33,9 +33,33 @@ export default class TreatmentVotingModalComponent extends Component {
...
-      'position'
-    );
+    const stemmingen = [];
+    const pageSize = 20;
+
+    const firstPage = yield this.store.query('stemming', {
+      'filter[behandeling-van-agendapunt][:id:]': this.args.behandeling.id,
+      sort: 'position',
+      page: {
+        size: pageSize,
```
lagartoverde, line 43: Wouldn't be easier to fetch with an absurd pageSize? like 999

lagartoverde (approved):
Looks good but it seems it should be easier to do

app/components/treatment/voting/modal.js
elpoelma: Yeah I'm still hesitating what the best approach is.

abeforgit (approved):

PR #354, opened 2022-11-03
Fix syntax issue that caused error when saving regulatory statement

elpoelma:
This PR fixes the method call syntax in order to save a regulatory statement which prevents an error.


lagartoverde (approved):
👍

PR #356, opened 2022-11-04
adapt query to the new model

lagartoverde:


app/services/regulatory-attachments-fetcher.js
```diff
@@ -17,20 +17,22 @@ export default class RegulatoryAttachmentsFetcher extends Service {
   *fetch() {
     const config = getOwner(this).resolveRegistration('config:environment');
     const sparqlQuery = `
+      PREFIX gn: <http://www.geonames.org/ontology#>
```
elpoelma, line 20: Why is this included?

app/services/regulatory-attachments-fetcher.js
```diff
@@ -17,20 +17,22 @@ export default class RegulatoryAttachmentsFetcher extends Service {
...
       PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
       PREFIX pav: <http://purl.org/pav/>
       PREFIX dct: <http://purl.org/dc/terms/>
       PREFIX schema: <http://schema.org/>
+      PREFIX gn: <http://data.lblod.info/vocabularies/gelinktnotuleren/>
       select distinct * where {
-        ?publishedContainer a ext:PublishedRegulatoryAttachmentContainer;
-          ext:currentVersion ?publishedAttachment.
-        ?publishedAttachment dct:title ?title.
-        ?publishedAttachment ext:content/mu:uuid ?fileId.
```
elpoelma, line 29: The fileId variable should still be requested right?

elpoelma (changes requested):
The template titles are correctly fetched, but it is currently not possible to instantiate a new document.

lagartoverde:
I don't know why I pushed this, as it's not finished, sorry my mind was in another place, I will set it as a draft

PR #361, opened 2022-11-08
When the notulen is published disable the editing

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-3640, I saw the issue and was simple so I implemented it, this has to be revisited if we implement the ability to republish a meeting, but for the moment it prevents confusion by the user.
@Dietr I used a simple design to show the user that a notulen has been published, just a pill on every heading, maybe you want to modify that
Screen Shot 2022-11-08 at 13 41 28


Dietr:
I would put a single pill in the header to notify that it's published. Then we don't need to repeat it on every title and it follows the article pattern.

lagartoverde:
In the header you mean next to the "Meeting scheduled..." text?

elpoelma (approved):
Works as expected, I agree with Dieter that we should move the `Notulen gepubliceerd` pill to the title.

PR #362, opened 2022-11-08
Feature/minor improvements

lagartoverde:
Fixed some problems I discovered on GN

https://user-images.githubusercontent.com/17006446/200581292-800c997a-58bb-4bab-9268-da6ebf990f4b.mov


nvdk (changes requested):
I don't agree with moving the queryparams to the edit route as done here, since it breaks app chrome on all the other subroutes (show, attachments and revisions)

lagartoverde:
> I don't agree with moving the queryparams to the edit route as done here, since it breaks app chrome on all the other subroutes (show, attachments and revisions)

yes sorry, my bad, I didn't even see it was already implemented on other routes, this problem was due to a bad merge I think

elpoelma (approved):
Works as expected

PR #363, opened 2022-11-08
Feature: prevent the duplicate inclusion of regulatory statements in a decision

elpoelma:
This PR does two things:
- It adds an `already attached` indication to regulatory statements which are already included.
- It disables the `insert` button when a statement is already coupled.

This PR solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3810


lagartoverde:
Not working for me, I will talk with you on Monday because I suspect that something is wrong with my setup

lagartoverde (approved):
Works good

PR #364, opened 2022-11-08
Disable 'insert statement' button if there is no besluit present

elpoelma:


lagartoverde (changes requested):
The button doesn't enable for me, not sure if I'm missing something

lagartoverde (approved):

PR #373, opened 2022-11-21
Feature: non editable published regulatory statements

elpoelma:
This PR ensures that published regulatory statements are not editable. It also ensures that logged in, read-only users dont see the editable view.

This PR solves https://binnenland.atlassian.net/browse/GN-3835?atlOrigin=eyJpIjoiYjAzMWJiOGM5ZTBhNGM5OGJkNmI1YWRlNDdlMWYwYWUiLCJwIjoiaiJ9.


lagartoverde (approved):
works good :)

PR #377, opened 2022-11-24
reset back to the url you were on after cancelling redirect

usrtim:
When cancelling the "are you sure you want to leave without saving" prompt when you click the browser back button, it should reset back to the url you were on.

https://user-images.githubusercontent.com/115715476/203875573-ce534561-dcf9-48b8-beb6-a5a548e2c62c.mp4

This fix is because of this issue: https://github.com/emberjs/ember.js/issues/18620 .

I also included a video, in which I tested every case I could think of.

### DIFF
--- a/app/components/confirm-route-leave.js
+++ b/app/components/confirm-route-leave.js
@@ -29,6 +29,10 @@ export default class ConfirmRouteLeaveComponent extends Component {
       this.args.onCancel(transition);
     } else {
       transition.abort();
+
+      if (window.history) {
+        window.history.forward();
+      }
     }
   }
 


app/components/confirm-route-leave.js
```diff
@@ -29,6 +29,10 @@ export default class ConfirmRouteLeaveComponent extends Component {
       this.args.onCancel(transition);
     } else {
       transition.abort();
+
+      if (window.history) {
```
nvdk, line 33: Ideally we solve this using the router, perhaps using https://api.emberjs.com/ember/4.8/classes/EmberRouter/methods/replaceWith?anchor=replaceWith to revert to the old route?

usrtim: I tried using **relaceWith** router method, but for some reason was not working. I will try again today and will let you know.

abeforgit: @nvdk replaceWith  is not available in ember 3.28, careful when you look at the api docs to select the right version

nvdk: hmm odd that it isn't offered, that seems like a bug :thinking:

nvdk: so yeah, tested it. the function is there, seems it somehow got dropped from the documentation

nvdk: that said, it seems non trivial to get this working with replaceWith, so I'd suggest to merge this as is

nvdk (approved):

app/components/confirm-route-leave.js
usrtim: I tried to get it working with **replaceWith**, but that is causing some weird behaviour.

PR #380, opened 2022-12-05
hide rs behind feature flag

nvdk:


usrtim (approved):

elpoelma (changes requested):
Works well overall, but it seems that the 'insert regulatory statement' functionality in an agendapoint is not yet disabled when needed.

nvdk:
> Works well overall, but it seems that the 'insert regulatory statement' functionality in an agendapoint is not yet disabled when needed.

thanks for catching that!

app/instance-initializers/features-setup.js
```diff
@@ -0,0 +1,17 @@
+import config from 'frontend-gelinkt-notuleren/config/environment';
+
+export function initialize(appInstance) {
+  const features = appInstance.lookup('service:features');
+  const featureConfig = config.featureFlags || {};
+  for (const key of Object.keys(featureConfig)) {
+    const val = featureConfig[key];
+    if (typeof val === 'string' || val instanceof String) {
```
lagartoverde, line 8: Do we need both checks? Is there any string that is not typeof string but instanceof String?

lagartoverde (changes requested):
Works well overall, as Elena said before me we also have to disable the regulatory statement plugin

app/instance-initializers/features-setup.js
lagartoverde: ok, then good job for catching that :)

PR #384, opened 2022-12-08
Feature/rs revisions

lagartoverde:


lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-3877 and https://binnenland.atlassian.net/browse/GN-3878

elpoelma (changes requested):
Seems to work as expected overall :smile: .
Some notes:
- the list of revisions shown in the dropdown does not seem to update when creating new versions unless you reload the page.
- There are still some inline styles used in the components

elpoelma (approved):

PR #386, opened 2022-12-09
Feature/create new rs versions

lagartoverde:


elpoelma (changes requested):
Seems to work well. We could maybe add a pill to each version in the dropdown to indicate if they have been published or not.

abeforgit (approved):
works well

PR #387, opened 2022-12-13
Feature/history modal

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-3880


elpoelma (changes requested):
The number of revisions show in the history modal is not always consistent, additionally it can be useful to show which revision is currently selected in the history modal.

elpoelma (approved):
Looks good now!

PR #388, opened 2022-12-14
Don't allow participant list insertion if the chairman is not present

lagartoverde:


app/components/participation-list/modal.hbs
```diff
@@ -5,6 +5,14 @@
         as |Modal|>
   <Modal.Body>
     <div class="au-o-flow">
+      {{#if this.error}}
```
abeforgit, line 8: I would put this error next to the "insert" button

atm when you are scrolled down, you don't see the error and it feels like the insert button just doesnt work

abeforgit (changes requested):
Just a smal QoL comment, works well otherwise
OK to merge for me, moving the alert is optional

abeforgit (changes requested):
come to think of it, wouldn't it be better to simply disable the present toggle if the mandatee = the chairman?

We have no backup flow for when they do set the chairman as absent, we simply don't allow them to proceed. So why even have the option in the first place?

lagartoverde:
I have made the changes you suggested, also kept the message and moved the error just in case, or if we want to add more errors to it

elpoelma (approved):
Works as expected.

PR #390, opened 2022-12-16
Make title error visible and wrap when content is long

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-2674

### DIFF
--- a/app/components/editor-document-title.hbs
+++ b/app/components/editor-document-title.hbs
@@ -1,5 +1,5 @@
 {{#if @readOnly}}
-  <h1 class="au-c-app-chrome__title">
+  <h1 class="au-c-app-chrome__title" title={{this.title}}>
     {{limit-content this.title 70}} 
   </h1>
 {{else}}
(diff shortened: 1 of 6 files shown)


app/components/editor-document-title.hbs
```diff
@@ -10,15 +10,27 @@
...
-          <input
-            class="au-c-app-chrome__title-input"
-            placeholder="Naamloos document"
-            id={{id}}
-            type="text"
-            value={{this.title}}
-            {{on "input" this.setTitle}}
-            {{did-insert this.focus}}
-          />
+          {{#if this.error}}
```
abeforgit, line 13: You can put the if statement inside the class value here

app/components/editor-document-title.hbs
```diff
@@ -10,15 +10,27 @@
...
+            <input
+              class="au-c-app-chrome__title-input au-c-input--error "
+              placeholder="Naamloos document"
```
abeforgit, line 16: i18n please

abeforgit (changes requested):

app/components/editor-document-title.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if @readOnly}}
   <h1 class="au-c-app-chrome__title">
-    {{this.title}}
+    {{limit-content this.title 70}}
```
abeforgit, line 3: probably want some way to see the full title when read-only. (when not read-only, you can see it by editing it)

abeforgit (changes requested):

elpoelma (approved):
Works as expected

PR #392, opened 2022-12-20
Feature/solve inconsistencies

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-3921 https://binnenland.atlassian.net/browse/GN-3922 and https://binnenland.atlassian.net/browse/GN-3923


elpoelma (changes requested):
- The status pill is correctly shown for each version
- There are still some issues when using the revisions as a read-only users, sometimes the page becomes blank and console errors pop up
- Read-only published and concept versions correctly show different messages
- When opening a published regulatory statement the banner with the text is not yet internationalized.

elpoelma (approved):
Works well now! :smile:

PR #393, opened 2022-12-20
hotfix: remove empty decisions from document on save

usrtim:
In order to prevent users from being able to create “faulty” documents by copy pasting etc, we would remove decisions that don’t have any content from the document when saving.

### DIFF
--- a/app/controllers/agendapoints/edit.js
+++ b/app/controllers/agendapoints/edit.js
@@ -93,14 +93,32 @@ export default class AgendapointsEditController extends Controller {
     } else {
       this.hasDocumentValidationErrors = false;
       const html = this.editor.htmlContent;
+      const cleanedHtml = this.removeEmptyDivs(html);
+
       const editorDocument =
         yield this.documentService.createEditorDocument.perform(
           this.editorDocument.title,
-          html,
+          cleanedHtml,
           this.documentContainer,
           this.editorDocument
         );
       this._editorDocument = editorDocument;
     }
+  }
+
+  removeEmptyDivs(html) {
+    const parserInstance = new DOMParser();
+    const parsedHtml = parserInstance.parseFromString(html, 'text/html');
+    const body = parsedHtml.body.childNodes;
+
+    for (const child of body) {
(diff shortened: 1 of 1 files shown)


app/controllers/agendapoints/edit.js
```diff
@@ -93,14 +93,32 @@ export default class AgendapointsEditController extends Controller {
...
     }
   }
+
+  removeEmptyDivs(html) {
+    const parserInstance = new DOMParser();
+    const parsedHtml = parserInstance.parseFromString(html, 'text/html');
+    const body = parsedHtml.body.childNodes;
+
+    for (const child of body) {
+      if (child.attributes?.typeof?.textContent.includes('besluit:Besluit')) {
```
nvdk, line 115: I would check for both the prefixed and full URI here.

app/controllers/agendapoints/edit.js
```diff
@@ -93,14 +93,32 @@ export default class AgendapointsEditController extends Controller {
...
+        if (child.textContent.trim() === '') {
+          child.remove();
+        }
+      }
+    }
+
+    return parsedHtml.documentElement.lastChild.innerHTML;
```
nvdk, line 122: shouldn't this just be `parsedHtml.body.innerHTML;` ?  That's more inline with how you start above

usrtim: yes, thanks for noticing that.

app/controllers/agendapoints/edit.js
usrtim: @nvdk please let me know if new commit address your concern ?

elpoelma (approved):
Works as expected. When removing all content of a decision (but keeping its div) it is correctly removed.

abeforgit (changes requested):
This only checks the top level children of the document. What if the div in question is further down the tree?
As we mentioned in the standup, there is https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll which will probably be useful here

PR #394, opened 2022-12-20
Pass sort to the backend on uittreksels publication table

lagartoverde:


app/routes/meetings/publish/uittreksels/index.js
```diff
@@ -21,6 +21,7 @@ export default class MeetingsPublishUittrekselsRoute extends Route {
     const query = {
       include: 'behandeling.versioned-behandeling',
       filter: { zitting: { ':id:': this.meetingId } },
+      sort: params.sort,
```
nvdk, line 24: we're also not passing page and size, best to add those as well. 
+ we might want to check if this is also an issue on other routes.

PR #397, opened 2022-12-26
Feature/support long titles in meeting screen

lagartoverde:
Basically if you create an agendapoint with a very long title in a meeting everything should work fine


elpoelma (approved):
Looks good, I am wondering if we also should add this shortening to the publication service... Probably not.

PR #400, opened 2023-01-10
Add case insensitive sorting to all text-based/title columns

elpoelma:
This PR ensures that most text-attributes (such as titles) are sorted in a case-insensitive way.
To be tested in combination with https://github.com/lblod/app-gelinkt-notuleren/pull/140.

Solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3912


abeforgit (approved):
works well!

usrtim (approved):

PR #402, opened 2023-01-17
Add new statuses for the regulatory statements

lagartoverde:
needs https://github.com/lblod/app-gelinkt-notuleren/pull/143


elpoelma (changes requested):
Works well overall, I suppose it could be useful to also show the status on the regulatory statement edit page.

elpoelma (approved):
Seems to work well for me now :)

PR #415, opened 2023-01-31
Addition of not-found route

elpoelma:
This PR does two things:
- Addition of an application-header component which either shows a login button, or the status of the currently logged in user, if there is one.
- Addition of a not-found route which matches either non-defined route path. It shows a message to a user and a link to the contact-page.

Partial solution to https://binnenland.atlassian.net/browse/GN-3989?atlOrigin=eyJpIjoiMmI3MmY1NmQxYWRkNDY4ODlkZTVmOGJkNGIyMGRjOWUiLCJwIjoiaiJ9.

### DIFF
--- a/app/components/application-header.hbs
+++ b/app/components/application-header.hbs
@@ -1,0 +1,21 @@
+
+
+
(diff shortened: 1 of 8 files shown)


lagartoverde (approved):
Looks good

PR #426, opened 2023-02-14
Internal/prevent hardcoded strings

lagartoverde:
Added the "no-bare-strings" rule to the linter and solve all the errors.
Disclaimer: this doesn't mean that everything in the app is translated as some strings are passed as arguments to components and also some of them are from the database, for the first one is still left as a TODO but this PR was getting too big.
Also I didn't translated anything of the legaal folder, I talked with Niels and this could cause problems.


translations/nl-BE.yaml
```diff
@@ -1,10 +1,29 @@
...
+  version-of: Versie van
+  no-new-versions: Geen nieuwere versies om weer te geven.
+  restore-version: Herstel versie en verwerp nieuwere versies
+  revisions-modal-title: Bent u zeker dat u deze versie wil herstellen?
+  published: Dit artikel is reeds gepubliceerd. Gepubliceerde agendapunten zijn niet editeerbaar.
+  saving: Bezig met opslaan
+  making-copy: Kopie aan het maken
+
+administrative-body-select:
+  period: "(period:"
```
elpoelma, line 21: I would extract the '(' from the translations.

elpoelma (changes requested):
Seems to be correct overall, I just fixed some small typos.

elpoelma (approved):
Fixed another typo and removed an unnecessary colon, but it should be ok to merge now.

PR #431, opened 2023-02-22
update ember-data to 3.28.13

lagartoverde:
Seems to solve the problems with the participation list, https://binnenland.atlassian.net/browse/GN-4118


elpoelma (changes requested):
Secretaries of a meeting are not correctly fetched. The following error is displayed: 
image

nvdk:
I think it would be better to await the functionaris higher up in the power select or make sure it's loaded in loadData

app/models/functionaris.js
```diff
@@ -1,6 +1,7 @@
 import Model, { attr, belongsTo } from '@ember-data/model';
+import MandatarisModel from './mandataris';
 
-export default class FunctionarisModel extends Model {
+export default class FunctionarisModel extends MandatarisModel {
```
abeforgit, line 4: why did you do this? and don't we need to set polymorphic on some relationships then?

app/components/participation-list.hbs
```diff
@@ -11,7 +11,7 @@
       {{/unless}}
     </li>
     <li class="au-c-list-divider__item">
-      {{t "participation-list.secretaris-label"}} <strong>{{this.secretary.isBestuurlijkeAliasVan.fullName}}</strong>
+      {{t "participation-list.secretaris-label"}} <strong></strong>
```
abeforgit, line 14: theres an empty strong tag here

package.json
```diff
@@ -66,7 +66,7 @@
     "ember-concurrency": "^2.3.7",
     "ember-config-helper": "^0.1.3",
     "ember-could-get-used-to-this": "^1.0.1",
-    "ember-data": "~3.27.1",
+    "ember-data": "~3.28.12",
```
abeforgit, line 69: could you bump to 3.28.13 while you're at it

abeforgit (changes requested):

lagartoverde:
@elpoelma is this finished? Because it looks good to me, and it works, so if it is, you can merge

elpoelma:
> I think it would be better to await the functionaris higher up in the power select or make sure it's loaded in loadData

I agree, I will try to look into this in this PR, or a future PR.

elpoelma:
> > I think it would be better to await the functionaris higher up in the power select or make sure it's loaded in loadData
> 
> I agree, I will try to look into this in this PR, or a future PR.

I honestly also don't see why the `await` would be needed, as the `functionaris`-models are already resolved in the `loadData` function.

abeforgit (changes requested):
- some problem with inverses when loading the people present at the meeting
image

kind of breaks the whole meeting page

elpoelma:
> * some problem with inverses when loading the people present at the meeting
>       image
> 
> 
> kind of breaks the whole meeting page

should be resolved now, see https://github.com/lblod/frontend-gelinkt-notuleren/pull/462#issuecomment-1562407050

abeforgit (changes requested):
getting this error when trying to open an agenda publication preview:
image

PR #434, opened 2023-03-03
Feature/implement new editor

lagartoverde:
Update editor to versions 3.2.1 and plugins to version 3.0.0


elpoelma (changes requested):
Seems to work as expected overall, I just noticed you forgot the template variable-card in angendapoints.

elpoelma (changes requested):
I also noticed you forgot to update the `zitting-text-document-container` component.

elpoelma (changes requested):
It seems that the intro/outro views of a meeting still have some buttons enabled which try to insert nodes which are not in their schema (such as links).

elpoelma (changes requested):
I discovered an error in the citation plugin: it repeatedly yields the following error:
`Uncaught TypeError: this.members[i] is undefined`. I will open a ticket for it and update this message.

elpoelma (approved):

PR #438, opened 2023-03-15
Ensure the meeting edit page shows the correct status

elpoelma:
This PR solves an issue with the displayed status on the meeting edit page. Whereas it showed `notulen gepubliceerd` before for both signed and published meetings, it now shows the correct status (signed when only signed and published when published).

This solves https://binnenland.atlassian.net/browse/GN-4135.


app/components/meeting-form.js
```diff
@@ -22,11 +23,43 @@ export default class MeetingForm extends Component {
   @service store;
   @service currentSession;
   @service router;
+  @service intl;
+
+  isPublished = trackedFunction(this, async () => {
+    await Promise.resolve();
```
lagartoverde, line 29: Why do we do this? I assume in order to avoid blocking, maybe we want to comment this little hack

lagartoverde (approved):
Looks good to me :)

app/components/meeting-form.js
abeforgit: I also don't know, and I would like to, before merging :)

PR #440, opened 2023-03-16
Only take latest versions of agendapoints into account when calculating the status of a regulatory statement

elpoelma:
Not a very nice solution, but this should ensure that the status of a regulatory statement is correctly shown. 
Should solve https://binnenland.atlassian.net/browse/GN-3995


abeforgit:
superceded by #443

PR #442, opened 2023-03-22
GN-4159 - Feature/create revoke signature button

lagartoverde:


lagartoverde:
Should solve https://binnenland.atlassian.net/browse/GN-4158 and https://binnenland.atlassian.net/browse/GN-4159, it needs the PR https://github.com/lblod/notulen-prepublish-service/pull/101 and https://github.com/lblod/app-gelinkt-notuleren/pull/150

app/components/publishing-log-hash.js
```diff
@@ -0,0 +1,16 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
+
+export default class PublishingLogHashComponent extends Component {
+  @tracked hash;
+
+  @task
+  *loadData() {
```
abeforgit, line 9: This breaks the fundamental ember contract of recalculating dependent data when their reactive dependencies change. This function depends on `this.args.log`, but it will not recalculate when that changes.

this is exactly _why_ ember-resources exist. 

Same thing in publishing-log-document-name

app/models/agenda.js
```diff
@@ -4,6 +4,7 @@ export default class Agenda extends Model {
   @attr agendaStatus;
   @attr agendaType;
   @attr renderedContent;
+  @attr('boolean', { defaultValue: false }) deleted;
```
abeforgit, line 7: why does an agenda have a deleted status?

app/models/versioned-agenda.js
```diff
@@ -4,6 +4,7 @@ export default class VersionedAgendaModel extends Model {
   @attr state;
   @attr content;
   @attr kind;
+  @attr deleted;
```
abeforgit, line 7: why does this one not have a defaultvalue?

abeforgit (changes requested):
why all the `model.get` calls? that method is deprecated in ED 4.x, and even before it was not "supposed" to be used.

not tested, since this PR broke the signing button, but that could be something with my setup

app/models/agenda.js
lagartoverde: Because the whole thing is a mess, the versionedAgenda is not linked to published/signed resources and is this one who it is

app/components/signatures/timeline-step.js
```diff
@@ -139,23 +142,85 @@ export default class SignaturesTimelineStep extends Component {
...
-    yield this.args.signing(signedId);
-    const signedResources = yield this.args.signedResources;
+    let signedResources = await this.args.signing(signedId);
+    if (!signedResources) {
+      signedResources = oldSignedResources;
+    }
     this.signedResources = signedResources.sortBy('createdOn');
+    const signedResource = signedResources.lastObject;
+    let versionedResource;
+    if (await signedResource.get('agenda')) {
```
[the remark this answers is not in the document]
lagartoverde: You are right, I wanted to simplify the code, but might lead to inconsistences, I will redo that

app/models/agenda.js
abeforgit: :+1:

app/routes/meetings/publish/publication-actions/detail.js
```diff
@@ -0,0 +1,20 @@
...
+      {
+        include: 'signed-resource,published-resource',
+      }
+    );
+    if (log.get('signedResource')) {
+      const signedResource = await log.get('signedResource');
+      return signedResource.get('content');
+    } else {
+      const publishedResource = await log.get('publishedResource');
+      return await publishedResource.get('content');
```
elpoelma, line 17: The usage of getters here can be replaced by just property accesses.

elpoelma, line 17: I don't think the await is necessary here.

app/controllers/meetings/publish/notulen.js
```diff
@@ -130,11 +132,19 @@ export default class MeetingsPublishNotulenController extends Controller {
       { method: 'POST' }
     );
     await this.muTask.waitForMuTaskTask.perform(taskId);
-    this.signedResources = await this.store.query('signed-resource', {
-      'filter[versioned-notulen][zitting][:id:]': this.model.id,
-      include: 'gebruiker',
-      sort: 'created-on',
+    await this.loadNotulen.perform();
+    const signedResources = this.notulen.signedResources;
```
elpoelma, line 136: Shouldn't this relation be awaited?

app/models/zitting.js
```diff
@@ -21,4 +21,5 @@ export default class ZittingModel extends Model {
   @hasMany('intermission', { inverse: 'zitting' }) intermissions;
   @hasMany('versioned-behandeling', { inverse: 'zitting' })
   versionedBehandelingen;
+  @hasMany('publishing-logs', { inverse: 'zitting' }) publishingLogs;
```
elpoelma, line 24: Shouldn't this be `publishing-log`?

elpoelma (commented):
I have to test this further, but I have some comments that you might want to look into :)

abeforgit (changes requested):
- signing besluitenlijst is still broken
- I don't particularly love the way this looks: 
image
(the different styles of brackets, inconsistent spacing, having "removed" that looks similar to a decision subtitle, etc)
- you can delete signatures from published documents

abeforgit:
A long overdue update:

This is now ready for review, as far as I can tell.
the following PR for the prepublishing service is required:
https://github.com/lblod/notulen-prepublish-service/pull/101


to include it in your stack, put the following in your override:
```yml
  preimporter:
    build:
      context: /the/location/of/the/repo/notulen-prepublish-service
      dockerfile: Dockerfile
    volumes:
      - /the/location/of/the/repo/notulen-prepublish-service:/app
```
and launch it with `docker-compose up -d --build`


What's changed:
- complete rework of the publish/uittreksels/show route, ditching the problematic timeline-step and the misguided `publish` service:
	- vastly simplified backend logic by putting it all in the controller, which made it clear how much unnecessary things were being done
	- actually using the route for data loading
	- extracted out the signing controls for reuse when we tackle the other routes, as that's the most complicated part that's still generic
- rework of the signing route to be jsonAPI compliant (thanks for the groundwork @nvdk) to avoid Ember Data caching issues. In the end I had to use query anyway, but it was still very useful

this fixes all the problems I found with signing the extracts, some of which were introduced by this pr, some of which were just always problematic cause the thing was held together with hopes and prayers

hopefully this can serve as a bit of a template to refactor the other publishing routes. 
It can also still be improved, there's quite a few easy UI-only components in the template waiting to be extracted, but I've had enough for now.

@nvdk would like your eyes mainly to check legal correctness with the deleting of signatures

abeforgit:
reported by niels in rocket chat:
- toggle of people present at meeting is broken
- people present at voting don't load
sign agenda as laarne writer, delete signature
    log out
    log in as laarne reader
    sign agenda
    
    -> error

abeforgit:
something I noted:
- deleting signature of notulen does not update the status pill

nvdk (changes requested):
Signing the notulen should not prevent the use from switching whether a certain decision is published (included) in the notulen or not. It looks like this is no longer possible on this branch. At least it requires a manual refresh by the user.

Further more it looks like the  signing takes the settings into account, where it was previously specified the whole document (all full decisions) should be signed at all times.

on master after signing:
Screenshot from 2023-06-01 22-04-44

on this branch after signing:
Screenshot from 2023-06-01 22-06-21

nvdk:
Also: After removing the second signature it still shows on the print preview of an uittreksel
Screenshot from 2023-06-01 22-12-44
Screenshot from 2023-06-01 22-13-19

app/components/publishing-log-document-name.js
```diff
@@ -0,0 +1,72 @@
...
+export default class PublishingLogDocumentNameComponent extends Component {
+  @tracked documentName;
+  @tracked deleted;
+  @tracked route;
+  @tracked behandeling;
+  @service intl;
+
+  data = trackedFunction(this, async () => {
+    const log = this.args.log;
+    await Promise.resolve();
```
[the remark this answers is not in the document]
nvdk: this is currently required due to an issue with embers render cycle and how resources work. not a 100% sure it applies here though.
Some background in https://github.com/NullVoxPopuli/ember-resources/issues/340 and https://github.com/emberjs/rfcs/issues/769

elpoelma:
> reported by niels in rocket chat:
> 
>     * toggle of people present at meeting is broken
> 
>     * people present at voting don't load
> 
>     * sign agenda as laarne writer, delete signature
>       log out
>       log in as laarne reader
>       sign agenda
>       -> error
> 
> 
> first two probably related to ember data still EDIT: can not reproduce the first two EDIT: also can't repro the last but awaiting more testing

I reproduced the second issue (people present at voting don't load) when upgrading to ember-data 4.8, but I fixed it in that PR. I could not reproduce the error on ember data 3.28.13. The main issue was that the relation of people present at the meeting was not correctly awaited.

elpoelma:
> Signing the notulen should not prevent the use from switching whether a certain decision is published (included) in the notulen or not. It looks like this is no longer possible on this branch. At least it requires a manual refresh by the user.
> 
> Further more it looks like the signing takes the settings into account, where it was previously specified the whole document (all full decisions) should be signed at all times.
> 
> on master after signing: Screenshot from 2023-06-01 22-04-44
> 
> on this branch after signing: Screenshot from 2023-06-01 22-06-21

I was also able to reproduce the same issue, it only shows the toggle again after a full manual refresh. Will check if it occurs on master too.

elpoelma (changes requested):
Seems to work well overall. The new publication actions page works as expected. 
I could also reproduce some issues @nvdk noticed, including:
- the `make public` toggles no longer show up after signing notulen, they only show up after a manual browser refresh

I had no issues with removing a signature, signing out, signing in and adding a signature again.

elpoelma:
> Seems to work well overall. The new publication actions page works as expected. I could also reproduce some issues @nvdk noticed, including:
> 
>     * the `make public` toggles no longer show up after signing notulen, they only show up after a manual browser refresh
> 
> 
> I had no issues with removing a signature, signing out, signing in and adding a signature again.

The issue with the toggles not being displayed is because the `Signatures::Notulen::BehandelingenList` component in the notulen route is not being rendered after the signature addition/removal. This is due to the `in-element` helper not fully working as expected. 

```hbs
{{#in-element this.containerElement}}
              <Signatures::Notulen::BehandelingenList
                @notulen={{this.notulen}}
                @prepublishedBehandelingen={{this.treatments}}
                @publicBehandelingUris={{this.publicBehandelingUris}}
                @toggle={{this.togglePublicationStatus}}
                @toggleAll={{this.toggleAllPublicationStatus}}
                @allBehandelingPublic={{this.allBehandelingPublic}}
              />
            {{/in-element}}
```
`this.containerElement` is correctly updated after the signature addition/removal, but the `in-element` helper does not pick this up and thus no longer renders the `Signatures::Notulen::BehandelingenList` component.

elpoelma:
> > Seems to work well overall. The new publication actions page works as expected. I could also reproduce some issues @nvdk noticed, including:
> > ```
> > * the `make public` toggles no longer show up after signing notulen, they only show up after a manual browser refresh
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I had no issues with removing a signature, signing out, signing in and adding a signature again.
> 
> The issue with the toggles not being displayed is because the `Signatures::Notulen::BehandelingenList` component in the notulen route is not being rendered after the signature addition/removal. This is due to the `in-element` helper not fully working as expected.
> 
> ```
> {{#in-element this.containerElement}}
>       <Signatures::Notulen::BehandelingenList
>         @notulen={{this.notulen}}
>         @prepublishedBehandelingen={{this.treatments}}
>         @publicBehandelingUris={{this.publicBehandelingUris}}
>         @toggle={{this.togglePublicationStatus}}
>         @toggleAll={{this.toggleAllPublicationStatus}}
>         @allBehandelingPublic={{this.allBehandelingPublic}}
>       />
> {{/in-element}}
> ```
> 
> `this.containerElement` is correctly updated after the signature addition/removal, but the `in-element` helper does not pick this up and thus no longer renders the `Signatures::Notulen::BehandelingenList` component.

The issues is that the `containerElement` property is not a tracked getter, so the `in-element` does not pick it up. Ideally, something like this: 
```js
get containerElement() {
    if (this.loadNotulen.isIdle) {
      return document.getElementById(this.behandelingContainerId);
    } else {
      return null;
    }
  }
```
should work. Not super clean, but it ensures that containerElement is tracked each time the loading state changes.

elpoelma:
> @elpoelma wouldn't just making `behandelingContainerId` tracked work as well?

The issue is that `behandelingContainerId` does not change.

nvdk:
ah right :sweat_smile:

abeforgit:
So your suggestion works @elpoelma but I have no idea why since that whole block is already in an if-block that conditionally renders on the loadNotulen idle state

elpoelma (changes requested):
Seems to work well now, not encountering technical issues any longer. Just a note, 
image
In the above page the margins of the back button are rather small, maybe we could increase it?

elpoelma (approved):
Good to go now :)

PR #449, opened 2023-04-11
Fix schema of zitting-text component

elpoelma:
This PR adds a few different missing marks to the `zitting-text-document-container` component. These missing marks were causing errors when opening the meeting `intro` or `outro` modals. The toolbar tried to display widgets which used marks which were not included in the schema. Maybe an idea: we could check for the existing of the node_spec/mark_spec a widget uses inside the widget itself, so it disables itself when the mark/node spec is not defined.

Additionally this PR converts the dynamic `schema` getter of the  `zitting-text-document-container` component to a normal property to prevent duplicate instantiation of the schema object. This also solves https://binnenland.atlassian.net/browse/GN-4206?atlOrigin=eyJpIjoiYjRjYTU2ZDhiMzYzNDgwNDgyY2UxNjMyNjFmOTA5ZDkiLCJwIjoiaiJ9.

### DIFF
--- a/app/components/zitting-text-document-container.js
+++ b/app/components/zitting-text-document-container.js
@@ -8,7 +8,11 @@ import {
   strikethrough,
   strong,
   underline,
+  subscript,
+  superscript,
 } from '@lblod/ember-rdfa-editor/plugins/text-style';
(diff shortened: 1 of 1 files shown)


lagartoverde (approved):
Looks good :)

PR #452, opened 2023-04-17
avoid saving editor document title without changes

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-4208, I couldn't reproduce the issue on local while I can easily do it on dev and qa, so it must have been fixed by other PR in the meantime.
I solved the issue where you could save a title even if it was unchanged

### DIFF
--- a/app/components/editor-document-title.hbs
+++ b/app/components/editor-document-title.hbs
@@ -26,6 +26,7 @@
           @skin="secondary"
           @icon="check"
           @hideText={{true}}
+          @disabled={{not this.titleModified}}
         >{{t "editor-document-title.change-title"}}</AuButton>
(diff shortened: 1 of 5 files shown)


elpoelma (approved):

abeforgit (changes requested):
the ticket also asks for an indication that it's been saved. So I'd suggest playing the same saving animation as for the document (but only IF it also saves the document)

lagartoverde:
I added a "Saved" AuPill next to the title, that hides itself after 30s to show the user the title got saved. Double check the dutch translation, I don't know if "Gered" is correct here, "Opgeslagen" was also suggested by Deepl

PR #453, opened 2023-04-20
Bugfix/problem with title in dev

lagartoverde:
In the end the problem with the title spotted by Kevin was indeed happening, but only in production with a Chrome browser, this solves it, basically the function toggle active was changing the active value of the component multiple times in the same computation because of the focus out event.

### DIFF
--- a/app/components/editor-document-title.hbs
+++ b/app/components/editor-document-title.hbs
@@ -33,7 +33,7 @@
   {{else}}
     <h1 class="au-c-app-chrome__title au-c-app-chrome__title--edit" title={{this.title}}>
       {{limit-content this.title 70}} 
-      <AuButton {{on "click" this.toggleActive}} @icon="pencil" @skin="tertiary" @hideText={{true}} >{{t "editor-document-title.change-title"}}</AuButton>
+      <AuButton {{on "click" this.enableEdit}} @icon="pencil" @skin="tertiary" @hideText={{true}} >{{t "editor-document-title.change-title"}}</AuButton>
     </h1>
     {{#if this.showSaved}}
       <AuPill
--- a/app/components/editor-document-title.js
+++ b/app/components/editor-document-title.js
@@ -41,7 +41,7 @@ export default class EditorDocumentTitleComponent extends Component {
   submit(event) {
     event.preventDefault();
     this.args.onSubmit?.(this.title);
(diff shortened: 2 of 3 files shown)


x-m-el (commented):
this code feels overly complex for what it does. If I see correctly:
`toggleActive` only gets used in template if active is false (=> should always set it to true)
in the code it always gets set (never toggled).

So toggleActive could be changed to 
```
setActive(val) {
  if (this.active && !this.title) {
      this.error = true;
    } else {
      this.active = val;
    }
}
```

x-m-el:
> > this code feels overly complex for what it does. If I see correctly: `toggleActive` only gets used in template if active is false (=> should always set it to true) in the code it always gets set (never toggled).
> > So toggleActive could be changed to
> > ```
> > setActive(val) {
> >   if (this.active && !this.title) {
> >       this.error = true;
> >     } else {
> >       this.active = val;
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With then also a change in the template to `{{on "click" (fn this.setActive true)}}`
> 
> Basically the problem is that if this.active is true it shouldn't be set to true again, because that causes some problems, I have now divided it into 2 functions so I can do the checks better, yesterday was too focus on making it work that I forgot to make it good :). Your solution is very similar to the one it was already there, the problem is that this.setActive(false) is called twice in the same computation and then it causes problems if you modify the this.active that your function does. Another solution would be to avoid the second call to the toggleActive but is called on a focusout event that I don't really know how to avoid

Wait, so setting isActive twice to the same value creates the problem? I wouldn't have guessed with the code :p
Wouldn't it then be nicer to solve this with [dedupeTracked](https://github.com/tracked-tools/tracked-toolbox#dedupetracked) then?
I think that function is meant for cases like this and it would avoid the extra checks and extra need for functions i think.

lagartoverde:
Tried with deduped tracked and the old code but the same error still appears

x-m-el (commented):
add a comment why the weird code is needed, so people don't refactor it away in the future and reintroduce the bug. If it's too hard to explain you could add a link to this PR (but that's often less clean). Then it can be merged :)

x-m-el (approved):

PR #458, opened 2023-05-03
Add chrome hacks plugin to GN

lagartoverde:


lagartoverde:
In draft until https://github.com/lblod/ember-rdfa-editor/pull/770 is merged and released

PR #459, opened 2023-05-04
GN-4240: fix Change agenda point title in zitting editor

dkozickis:
https://binnenland.atlassian.net/browse/GN-4240


app/controllers/agendapoints/edit.js
```diff
@@ -265,14 +265,30 @@ export default class AgendapointsEditController extends Controller {
...
       await this.documentService.createEditorDocument.perform(
         title,
         html,
         this.documentContainer,
         this.editorDocument
       );
+
     this._editorDocument = editorDocument;
+
+    await behandeling.save();
```
lagartoverde, line 291: you are saving the behandeling but not the agendaitem so the changes are not persisted

lagartoverde (changes requested):
What does it happen if the document is not linked to any behandeling, in that case this function will fail and not update the title.
Also we are reaching the point where updating the title is becoming too slow, I would update the title on the editor document the first thing so the user doesn't notice or show a loader somewhere, currently the UX is weird

PR #461, opened 2023-05-12
Always show time on relative dates

lagartoverde:


lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-4213

PR #473, opened 2023-05-30
GN-4311: Bump `@lblod/ember-rdfa-editor` package

dkozickis:
- Bump `@lblod/ember-rdfa-editor` package to fix annotation not present for some structures

https://binnenland.atlassian.net/browse/GN-4311


lagartoverde (approved):
You can automatically merge these yourself unless is a breaking release and you want additional testing

PR #475, opened 2023-06-12
Fix agendapoint title saving

elpoelma:
This ensures that the title of an agendapoint document is saved correctly after editing it. Additionally, if the agendapoint document is added to a meeting (and thus has a treatment), the title of the subject (onderwerp) of the treatment is also updated.

Fixes both https://binnenland.atlassian.net/browse/GN-4349 and https://binnenland.atlassian.net/browse/GN-4240


abeforgit (approved):
works well, tried a bunch of scenarios

PR #498, opened 2023-06-20
Fix - Add publication action when deleting signature of document

x-m-el:
### Overview
When any user deleted a document signature, it would not show in the history tab (publication-actions). This was because `this.meeting` was used but that isn't defined. 

- [GN-4355: Actions for the mock-user (role 'Lezer') aren't being logged under 'Publication actions'](https://binnenland.atlassian.net/browse/GN-4355) => note that this PR is a fix for all users, not only "reader role".
- https://github.com/lblod/app-gelinkt-notuleren/pull/163 => Note: you can test this PR separately for a writer-role. To also test this with a reader-role, you'll need this backend.
- also fixes the following ticket (which was added after creation of this pr): https://binnenland.atlassian.net/browse/GN-4385

### How to test/reproduce
1. start with an old setup (e.g. master)
2. mock-login to a writer-role ("schrijver")
3. sign a document. Do this by creating a meeting, filling in the required fields (at least one agenda point) and pressing the right top button "Sign and Publish" (only appears if everything required is done for the meeting!)
4. Go to publication actions. The signing of the user will not be present.

To test the fix, follow the exact same procedure.
In step `4.` the sign action will now appear in the list.

### Challenges/uncertainties
`this.notulen` has the data needed (`.zitting`), but I am not familiar enough with the data structure to know if this is (always) the correct place to get this info.


lagartoverde (approved):
Works for me 👍

x-m-el:
update: another linked ticket fixed with this PR: https://binnenland.atlassian.net/browse/GN-4385

x-m-el:
Also tested if publishing gets recorded, those all seem to be in order :+1:

PR #499, opened 2023-06-22
GN-4362 fix signatures disappearing after publishing

abeforgit:
# Overview

For notulen, the VersionedNotulen you sign is not the same as the one
you publish. So I had to revert my earlier change to go back to saving
any non-empty signedresource relationships.

### connected issues and PRs:

<https://binnenland.atlassian.net/browse/GN-4362> 
### Setup

# How to test/reproduce

See ticket, make meeting, sign, then publish before, signature
dissapeared now, it doesnt 
### Challenges/uncertainties

# Checks PR readiness

-   [X] UI: works on smaller screen sizes
-   [X] UI: feedback for any loading/error states
-   [X] Check cancel/go-back flows
-   [X] Check database state correct when deleting/updating (especially
    regarding relationships)
-   [X] changelog
-   [X] npm lint
-   [X] no new deprecations

### DIFF
--- a/app/controllers/meetings/publish/notulen.js
+++ b/app/controllers/meetings/publish/notulen.js
@@ -103,13 +103,16 @@ export default class MeetingsPublishNotulenController extends Controller {
             this.notulen = notulen;
             notulenSet = true;
           }
-          this.signedResources = signedResources.toArray();
-          if (!notulenSet) {
-            this.notulen = notulen;
+          if (signedResources.length) {
+            this.signedResources = signedResources.slice();
+            if (!notulenSet) {
+              this.notulen = notulen;
+            }
           }
         })
       );
     } else {
+      console.log("creating new versioned notulen");
       try {
         const { content, errors } =
           await this.createPrePublishedResource.perform();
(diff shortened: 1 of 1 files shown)


abeforgit:
> It works as expected +1
> 
> I can't really grasp the code though. Is this correct:
> 
>     * first there is no versionedNotulen. a dummy versionedNotulen record is created to show instead.
> 
>     * a user can then sign, which will create a signed resource (and also a versionedNotulen in the backend? What happens for a second signature?)
> 
>     * Can publish, which also creates a versionedNotulen (with a publishedResource attached).
> 
> 
> In loadNotulen:
> 
>     1. if published, that notulen is always used
> 
>     2. if no published but a signed notulen, that signed notulen is used
>        Is there always only one signed notulen (with one or two signatures)?
> 
>     3. if no signing/publish, a "dummy" notulen record is used
> 
> 
> I think `notulenSet` would better be named `publishedNotulenSet` to make it clear why it shouldn't be set to true when set for a signedResource.

sounds about right but indeed the logic is completely bonkers
not gonna refactor it now though not enough time

PR #509, opened 2023-06-26
GN-4358: addition of a document-type filter to the publication actions page

elpoelma:
### Overview
This PR adds a dropdown component to the publication actions page which allows users to filter on document type.
The values a user can select between are:
- Agenda
- Decision list
- Extract
- Document/Notulen

Jira ticket: https://binnenland.atlassian.net/browse/GN-4358?atlOrigin=eyJpIjoiOTViNmRjMzcwN2ExNGJkY2E5YTMyZTNkZDc2MzY1ODMiLCJwIjoiaiJ9
Design: https://www.figma.com/file/pW370BHe4ES5eN0vHhArA5/GN-4022?type=design&node-id=0-1&t=Ncms3NPhRgxXECZW-0

### How to test/reproduce
Visit the publication actions page and select a document type using the dropdown. Only the selected document type should be shown and the pagination should reset to page 0.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/app/components/publishing-log-document-name.js
+++ b/app/components/publishing-log-document-name.js
@@ -60,7 +60,7 @@ export default class PublishingLogDocumentNameComponent extends Component {
           const versionedBehandeling = logResource.versionedBehandeling;
           if (versionedBehandeling) {
             documentName = this.intl.t(
-              'meetings.publish.publication-actions.treatment'
+              'meetings.publish.publication-actions.extract'
             );
             versionedResource = await logResource.versionedBehandeling;
             const behandeling = await versionedResource.behandeling;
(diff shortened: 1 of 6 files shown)


abeforgit:
I know I'm being annoying but please stick to the jira format, so GN-4358 (will make parsing easier later)

lagartoverde (changes requested):
The filtering is kept evern if you change meetings which can be confusing for the user because he can lose track of what is being filtered, apart from that everything else looks good 👍

elpoelma:
> The filtering is kept evern if you change meetings which can be confusing for the user because he can lose track of what is being filtered, apart from that everything else looks good +1

Should be solved with the `resetController` hook.

abeforgit (approved):

PR #513, opened 2023-06-29
GN-4360: Unable to edit meeting after reverting signatures + no alert

dkozickis:
#### Connected issues and PRs:
https://binnenland.atlassian.net/browse/GN-4360

### Setup
Usual setup, nothing special

### How to test/reproduce
You need to have a Notulen ready to be signed.

1. Sign with one user
2. Observer that you cannot edit the meeting anymore
3. Sign with second user
4. Remove the second user signature
5. Observe the notification
6. Observer that you cannot edit the meeting still
7. Remove the first signature
8. Observe there's no more notification
9. Observer that you now _can edit the meetinf_

See video at - https://share.cleanshot.com/JJVHDGlj

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


app/components/meeting-form.js
```diff
@@ -38,12 +38,19 @@ export default class MeetingForm extends Component {
...
     // This is needed here, see https://github.com/NullVoxPopuli/ember-resources/issues/340
     await Promise.resolve();
-    const publishedNotulen = await this.store.query('versioned-notulen', {
-      'filter[zitting][id]': this.zitting.get('id'),
-      'filter[:has:signed-resources]': 'yes',
-      'fields[versioned-notulen]': 'id',
+
+    const signedResources = await this.store.query('signed-resource', {
+      'filter[versioned-notulen][zitting][:id:]': this.zitting.get('id'),
+      'filter[:has-no:deleted]': 'yes',
```
abeforgit, line 44: this means you will only get signed resources that don't have the "deleted" field
If they have the field but it is false, you won't get them
```
        'filter[:or:][deleted]': false,
        'filter[:or:][:has-no:deleted]': 'yes',
```
afaik this is the way to get both

app/components/meeting-form.js
```diff
@@ -38,12 +38,19 @@ export default class MeetingForm extends Component {
...
     });
-    return !!publishedNotulen.firstObject;
+
+    const arraySignedResources = signedResources.toArray();
+
+    if (!arraySignedResources.length) {
+      return false;
+    }
+
+    return arraySignedResources.toArray().some((resource) => !resource.deleted);
```
abeforgit, line 53: duplicate toArray here

abeforgit, line 53: and given the filter in the query, the some() should also not be needed

app/controllers/meetings/publish/notulen.js
```diff
@@ -99,29 +100,45 @@ export default class MeetingsPublishNotulenController extends Controller {
     return !!this.publishedResource;
   }
 
+  async loadSignedResources(versionedNotulenId) {
+    const signedResources = await this.store.query('signed-resource', {
```
abeforgit, line 104: I would do the filter in the query

It's almost impossible to go wrong _in this specific scenario_ cause you can't really create more than 2 deleted signatures in sequence

but if there _would_ be `>pagesize` (aka more than 20) deleted signed resources that were newer than the last undeleted one, you would get no undeleted resources in your result.

...which is pretty much impossible here unless something goes quite badly wrong. But still, it's good practice to filter on the backend

abeforgit (changes requested):
works afaict, some small code comments

elpoelma (approved):
Works as expected and the code seems pretty sane to me :)

abeforgit (approved):

PR #545, opened 2023-08-02
Hide template comments in all previews

x-m-el:
### Overview
Hide template comments when previewing documents (but keep showing when editing/filling them in). 
Template comments are only needed to be shown when editing a document.

##### connected issues and PRs:
This PR needs #541 which updates the editor packages. Note that this PR is a PR to master, split from #541, which means it also includes the #541 changes. The changes for hiding template comments can all be found in `app/styles/project/_c-editor-preview.scss`
[GN-4441](https://binnenland.atlassian.net/browse/GN-4441)
Changes in the publishing service: https://github.com/lblod/notulen-prepublish-service/pull/105

### Setup
1. checkout https://github.com/lblod/notulen-prepublish-service/pull/105 and add it as your preimporter in e.g. docker-compose.override.yml of app-gelinkt-notuleren. See https://github.com/lblod/notulen-prepublish-service#development-and-testing
2. run frontend (ember serve --proxy ...)

### How to test/reproduce
Make sure all template comments are hidden (except for where editing the document).
To easily add them do a document, add 
```
<TemplateCommentsPlugin::Insert @controller={{this.controller}}/>
```
to an editor. 

### Challenges/uncertainties
Hiding is done via css, which is the most easy part (and quickest, as there is no document changing).
It was added to all "preview" classes, because it is unclear where they might be used and end up in. So safest is the always hide them.

PR #552, opened 2023-08-08
GN-4264: enable template comments in GN

abeforgit:
### Overview

bumps plugins to 10.0.0
updates config to reflect the interface changes
removes outdated references to templateVariablePlugin from environment.js
enables template-comments-plugin on both editor configs

[GN-4264](https://binnenland.atlassian.net/browse/GN-4264)

### Setup

just run it

### How to test/reproduce

check template plugin, and variable plugins
### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [ ] no new deprecations

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,7 +85,7 @@
         "qunit": "^2.19.2",
         "qunit-dom": "^2.0.0",
         "release-it": "^16.1.3",
-        "sass": "^1.64.1",
+        "sass": "^1.64.2",
         "tracked-built-ins": "^3.1.1",
         "uuid": "^9.0.0",
         "webpack": "^5.88.0"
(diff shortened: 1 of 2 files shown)


x-m-el:
Probably want to "wait" on [PR to hide template comments when publishing](https://github.com/lblod/notulen-prepublish-service/pull/105) before merging this? So we don't get template comments in published data.

Both [PR for hiding template comments in frontend GN](https://github.com/lblod/frontend-gelinkt-notuleren/pull/545) and here template comments are added to editor, so will need to look out for at merging.

Is there any clear indication if we want template comments to be insertable in GN, like in agendapoints? Because I'd think there isn't more example/info to be given at this stage. Or are they meant as any general "extra information" (so an actual _comment_ :p ), to give some information about what to do in a certain agenda point?

Note that at this point they are hidden when previewing an agenda and are only shown during editing. So as far as I can see, the use would only be if an agenda point is created/edited in multiple stages and not at once.

x-m-el (changes requested):
still have to import `@import "template-comments-plugin";` in app styles (app.scss). Note: was missing from documentation to add this, this is added in a PR for template comment readme update.

For the rest is good. Tested comments and variables in agenda and regulatory statement :+1:

abeforgit:
Yeah I considered waiting, but given that the corresponding feature is already active on GN QA, if templates containing them will be used now in GN they will simply render as flat text, which I though was about as bad as them showing up in publications, but this at least can get the testers going

I was also debating whether or not GN users had to be able to insert them or not, figured why not. From what I gather the "edit in multiple stages" workflow is a real usecase, and a finished regulation/agenda item can be completed by different people. 

And on principle I want to move away from these "if I delete this, I can't put it back" scenarios where its there in the template, but the GN user lacks the tools to make that "thing" they see. They can already edit the comments anyway, which I'm completely fine with as well.

Hidden on preview is also perfect imo, the reader never needs to see them, only the person editing the document should.

And thanks for the hint about the css, I almost made a ticket for cleaning up the styling :see_no_evil:

abeforgit:
> so will need to look out for at merging

merge conflicts are like spiders: they're more afraid of you than you are of them

PR #578, opened 2023-09-19
GN-4266: Reference published decisions

dkozickis:
### Overview

Bumps "plugins" to enable published decisions insert

CleanShot 2023-09-19 at 12 27 30@2x

#### Connected issues and PRs:

https://binnenland.atlassian.net/browse/GN-4266

### Setup

1. Checkout
2. `ember s --proxy https://dev.gelinkt-notuleren.lblod.info`

### How to test/reproduce

Check whether it is possible to insert reference for a "Besluit". `decisionsEndpoint` might have to be changed to https://publicatie.gelinkt-notuleren.lblod.info/sparql as the production does not have `/sparql` yet.

### Challenges/uncertainties

Production does not have `/sparql` yet

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (changes requested):
Fetching decisions using the citation plugin does not work at the moment because it tries send a sparql query to https://publicatie.gelinkt-notuleren.vlaanderen.be/sparql. At the moment, GN publication production does not have a sparql endpoint yet, so the request fails. 

I would make the `decisionsEndpoint` configurable with the ember ENV, so we can configurable it using environment variables in docker-compose. This ensures we can point DEV to DEV, QA to QA and PROD to PROD. A similar approach is taken for the roadsign-regulation plugin. For development builds, I would set `decisionsEndpoint` to https://publicatie.dev.gelinkt-notuleren.lblod.info/sparql.

Additionally, I would also add changesets describing the editor and editor-plugins bump, as well as the `ember-modifier` downgrade.

PR #582, opened 2023-09-29
Also save title when saving document + fix save pill and don't allow empty titles

x-m-el:
### Overview
See mostly https://github.com/lblod/frontend-reglementaire-bijlage/pull/191 (copied from there)

- Empty titles are not permitted
- If title is empty, an error shows up
- save pill is only shown if something is actually saved
- save pill logic is a lot simpler and more logical, using the state of ember-concurrency, instead of keeping the exact same state ourselves.
- title is automatically save if focus is lost. => **ticket fix**: this fixes the original ticket issue. When pressing save-document button, the title input box will lose focus, making it automatically save together with saving the document.

[binnenland.atlassian.net/browse/GN-4473](https://binnenland.atlassian.net/browse/GN-4473)

similar changes in https://github.com/lblod/frontend-reglementaire-bijlage/pull/191

### How to test/reproduce
- try to save the title of any document (e.g. a regulatory statement). See if it works as expected and save-pill does not show up if nothing was saved.
- It should never be possible to save an empty title.
- Pressing save for the document should also save the title.
- Press escape and see your title not being changed.

### DIFF
--- a/.changeset/wet-forks-smell.md
+++ b/.changeset/wet-forks-smell.md
@@ -1,0 +1,8 @@
+---
+'frontend-gelinkt-notuleren': minor
+---
+
+- automatically save the document title if clicking outside of the title input box
+- Do not allow empty document titles
+- saving a document will also save its title
+- the "saved" pill will only show up if something is actually saved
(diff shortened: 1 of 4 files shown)


elpoelma (approved):
Very nice improvements :)

PR #583, opened 2023-09-29
GN-4525 Changes to support removal of ember-data from plugins

piemonkey:
### Overview
Pull standard-template-service from plugins and pass templates in to the standard template plugin.
Edit: modified to also pass in classificatieUri to besluit-type-plugin as needed by [plugins](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/314)

https://binnenland.atlassian.net/browse/GN-4525
https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/313
https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/314

### Setup
Needs the latest plugins as that uses the passed in templates correctly. This shouldn't break with older versions of the plugins though as it's simply passing in an arg that isn't used.

### How to test/reproduce
Test that editing agenda points and inserting templates works as expected.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/two-pants-thank.md
+++ b/.changeset/two-pants-thank.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Pass standard templates into standard template plugin instead of it loading them itself
--- a/app/components/document-creator.js
+++ b/app/components/document-creator.js
@@ -15,7 +15,6 @@ export default class DocumentCreatorComponent extends Component {
   @tracked errorSaving;
 
   @service store;
-  @service standardTemplatePlugin;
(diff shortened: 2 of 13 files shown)


abeforgit (changes requested):
the flow of making a document from an empty template, and then inserting a template, is broken:
image
image

piemonkey:
Ah, turns out the `reload` of templates was in fact doing something. @abeforgit I've fixed that now.

elpoelma (approved):
Seems to work well now :)

PR #586, opened 2023-10-03
GN-4505 Use instantiateUuids util from standard-template-plugin

piemonkey:
### Overview
Enables support for `generateBoundUuid()` from the updated version of that util. This allows for matching uris in the template to remain matching in the document.
Will need a new plugin release before merging.

https://binnenland.atlassian.net/browse/GN-4505
Requires: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/315
Required by: https://github.com/lblod/frontend-reglementaire-bijlage/pull/192

### Setup
Needs plugins from [the relevant PR](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/315) for the util to be exported.

### How to test/reproduce
Backwards compat can be tested by instantiating a regulatory statement and inspecting to verify uuids.
New features requires a template with `generateBoundUuid(<some argument>)` template substitutions.

### Challenges/uncertainties
It was unclear whether having separate generateUuids and generateBoundUuids functions was preferable to having just one function with an optional argument. Switching to just one overloaded function is trivial, but it's unclear if the current generateBoundUuids is used elsewhere already or not.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit:
@piemonkey could you bump to the latest plugins and then its good to merge

PR #591, opened 2023-10-06
GN-4529: ensure editor is only initialized after prefix attribute of wrapping div has been set

elpoelma:
### Overview
This PR ensures that the editor component is only initialized after the prefix attribute of it's wrapping container has been set.

This PR:
- introduces a `setUpPrefixAttr` modifier on the wrapping container which sets the prefix attribute. It also controls the `ready` property which indicates when the editor may be initialized.
 * After the `prefix` attribute has been set, the `ready` property is set to `true`
 * When the element linked to the `setUpPrefixAttr` modifier is destroyed, the `ready` property is set to `false` (as the prefix attribute no longer exists)
=> this also means that the `setUpPrefixAttr` modifier is not generic and is only meant to be used in combination with the wrapping container.

This PR should solve the issue with the datastore calculation after having saved a document. (When the datastore is calculated, the `prefix` attribute should be set)

Should solve [GN-4529](https://binnenland.atlassian.net/browse/GN-4529)

### How to test/reproduce
- Create an empty agendapoint
- Notice that you have the option to insert a template
- Save the document
- The options to insert templates should still be there.

### Challenges/uncertainties
- The modifier to set the prefix attribute is necessary as ember does not allow to set the `prefix` attribute directly in a template (see https://github.com/emberjs/ember.js/issues/19369)
- The new `setUpPrefixAttr` is not generic and should only be used once (as it controls the component property `ready`). I could not find a better way to solve this.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/.changeset/famous-balloons-smell.md
+++ b/.changeset/famous-balloons-smell.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Ensure editor is only initalized after correct prefix attribute has been set
--- a/app/components/rdfa-editor-container.hbs
+++ b/app/components/rdfa-editor-container.hbs
@@ -24,7 +24,7 @@
     </div>
   </AuBodyContainer>
 {{else}}
-  <AuBodyContainer vocab="{{this.vocab}}" {{did-insert this.setPrefix}}
+  <AuBodyContainer vocab="{{this.vocab}}" {{this.setUpPrefixAttr}}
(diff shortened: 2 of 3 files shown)


elpoelma:
> Huh, I don't get why this wasn't working with `did-insert` as I thought `EditorContainer` should only be initialised once `this.ready` is set to true?
> 
> Anyway, looks good.

Yes, but the `this.ready` attribute was never reset to false when saving.

PR #606, opened 2023-11-16
 GN-XXX: Fix agendapoint/regulatory attachment template "reload"

dkozickis:
### Overview

```ts
 /**
       * Document Creator component is used by two different screens:
       *   - Create agenda point flow
       *   - Create regulatory statements flow
       *
       * `templates` coming from regulatory statements flow are _NOT_ `TemplateModel` instances,
       * they are just plain objects with a `title` and a `loadTemplateBody` property. So we
       * have to use `loadTemplateBody` to load the template body if we need to build the template
       * when creating a regulatory statement.
       *
       * This was previously checking and calling `this.template.reload`, but that was causing
       * `reload` to mutate from function to a "boolean" when `TemplateModel` of `ember-data` is used,
       * causing errors when calling `reload` on the template again, as it became a boolean.
       *
       * The fix was to change `reload` to `loadTemplateBody` in `RegulatoryAttachmentsFetcher`
       */
```

### How to test/reproduce

Below bug should not occur again

* Create Agenda Point - workds fine
* Go back to list
* Create another: "error: template.reload is not a function"
* Refresh
* Create Agenda Point - works again

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/.changeset/wild-melons-dress.md
+++ b/.changeset/wild-melons-dress.md
@@ -1,0 +1,5 @@
+---
+"frontend-gelinkt-notuleren": patch
+---
+
+GN-XXX: Fix agendapoint/regulatory attachment template "reload"
--- a/app/components/document-creator.js
+++ b/app/components/document-creator.js
@@ -79,9 +79,24 @@ export default class DocumentCreatorComponent extends Component {
 
   async buildTemplate() {
     if (this.template) {
-      if (this.template.reload) {
-        // regular templates from templatesForContext do not return body of template
-        await this.template.reload(this.template);
+      /**
+       * Document Creator component is used by two different screens:
+       *   - Create agenda point flow
+       *   - Create regulatory statements flow
+       *
+       * `templates` coming from regulatory statements flow are _NOT_ `TemplateModel` instances,
+       * they are just plain objects with a `title` and a `loadTemplateBody` property. So we
+       * have to use `loadTemplateBody` to load the template body if we need to build the template
+       * when creating a regulatory statement.
+       *
+       * This was previously checking and calling `this.template.reload`, but that was causing
+       * `reload` to mutate from function to a "boolean" when `TemplateModel` of `ember-data` is used,
+       * causing errors when calling `reload` on the template again, as it became a boolean.
+       *
+       * The fix was to change `reload` to `loadTemplateBody` in `RegulatoryAttachmentsFetcher`
+       */
+      if (this.template.loadTemplateBody) {
+        await this.template.loadTemplateBody(this.template);
       }
       const trimmedHtml = this.template.body.replace(/>\s+</g, '><');
       return instantiateUuids(trimmedHtml);
--- a/app/services/regulatory-attachments-fetcher.js
+++ b/app/services/regulatory-attachments-fetcher.js
@@ -27,7 +27,7 @@ export default class RegulatoryAttachmentsFetcher extends Service {
           pav:hasCurrentVersion ?container.
         ?container dct:title ?title;
           mu:uuid ?fileId.
-        OPTIONAL { 
+        OPTIONAL {
           ?container schema:validThrough ?validThrough.
         }
         FILTER( ! BOUND(?validThrough) || ?validThrough > NOW())
@@ -56,7 +56,7 @@ export default class RegulatoryAttachmentsFetcher extends Service {
       const bindings = json.results.bindings;
       const templates = bindings.map((binding) => ({
         title: binding.title.value,
-        reload: async (template) => {
+        loadTemplateBody: async (template) => {
           const response = await fetch(
             `${config.regulatoryStatementFileEndpoint}/${binding.fileId.value}/download`,
           );


abeforgit:
Woah thats wild, make sure to double check any places where templates are used including creating RS and creatinf APs from meeting screen

elpoelma:
After looking into a bit _why_ the `reload` property was turned into a boolean, I discovered this piece of recent code (include in the ember-data 4.12 update): https://github.com/emberjs/data/blame/b176861834bc47da233b8fafe57d95a5c5d197fc/packages/model/src/-private/model-methods.ts#L60-L61

elpoelma:
The issue is that we need to support two type of reload methods with different signatures:
- Regulatory statement templates contain a `reload` method that expects a `Template` object
- Agendapoint templates (which are `ember-data` models) contains a `reload` method that expect an optional `AdapterOptions` object.

The issue is that we were also passing the template object to the `ember-data` reload method, which is of course not supported, and the `reload` property was mutated to `true` due to recent code changes.

This still gives us the following issue: we should be able to call the `reload` method of both type of templates. Or: we could indeed rename the method on the regulatory statement template objects, and still call reload (without arguments) if it exists.

elpoelma:
Also noticed that since the recent migration of the `standard-template` service, we are doing a `findAll` (without fields) when fetching the agendapoint templates. This seems unnecessary to me, as it also fetches the content of the agendapoint templates when it is not necessary: we should only fetch the content of a template when we really need it.

@dkozickis @piemonkey What are your thoughts on this?

elpoelma:
Proposal of update code:

```js
import Service, { service } from '@ember/service';
import { tracked } from '@glimmer/tracking';
import { task } from 'ember-concurrency';
import { getOwner } from '@ember/application';

class RegulatoryStatementTemplate {
  title;
  fileId;
  endpoint;
  body;
  constructor({ title, fileId, endpoint }) {
    this.title = title;
    this.fileId = fileId;
    this.endpoint = endpoint;
  }

  async reload() {
    const response = await fetch(
      `${this.endpoint}/${this.fileId.value}/download`,
    );
    this.body = await response.text();
  }
}
export default class RegulatoryAttachmentsFetcher extends Service {
  @service session;
  @service store;

  @tracked account;
  @tracked user;
  @tracked group;
  @tracked roles = [];

  fetch = task(async () => {
    const config = getOwner(this).resolveRegistration('config:environment');
    const sparqlQuery = `
      PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
      PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
      PREFIX pav: <http://purl.org/pav/>
      PREFIX dct: <http://purl.org/dc/terms/>
      PREFIX schema: <http://schema.org/>
      PREFIX gn: <http://data.lblod.info/vocabularies/gelinktnotuleren/>
      select distinct * where {
        ?publishedContainer a gn:ReglementaireBijlageTemplate;
          mu:uuid ?uuid;
          pav:hasCurrentVersion ?container.
        ?container dct:title ?title;
          mu:uuid ?fileId.
        OPTIONAL {
          ?container schema:validThrough ?validThrough.
        }
        FILTER( ! BOUND(?validThrough) || ?validThrough > NOW())
      }
    `;
    const details = {
      query: sparqlQuery,
      format: 'application/json',
    };
    let formBody = [];
    for (const property in details) {
      const encodedKey = encodeURIComponent(property);
      const encodedValue = encodeURIComponent(details[property]);
      formBody.push(encodedKey + '=' + encodedValue);
    }
    formBody = formBody.join('&');
    const response = await fetch(config.regulatoryStatementEndpoint, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
      },
      body: formBody,
    });
    if (response.status === 200) {
      const json = await response.json();
      const bindings = json.results.bindings;
      // const templates = bindings.map((binding) => ({
      //   title: binding.title.value,
      //   reload: async (template) => {
      //     const response = await fetch(
      //       `${config.regulatoryStatementFileEndpoint}/${binding.fileId.value}/download`,
      //     );
      //     template.body = await response.text();
      //   },
      // }));
      const templates = bindings.map(
        (binding) =>
          new RegulatoryStatementTemplate({
            title: binding.title.value,
            fileId: binding.fileId.value,
            endpoint: config.regulatoryStatementFileEndpoint,
          }),
      );
      return templates;
    } else {
      return [];
    }
  });
}
```

piemonkey:
@elpoelma It depends on the size of the agenda point contents, but yeah, if they aren't small, then we should only ask for the fields we need. Actually, now that I think of it, I'm maybe thinking in relational DB terms, where adding a field only costs as much as the time to read it and send it, but I guess in a triplestore it also adds lookup time, so we should always limit the fields we ask for if we don't need them all.

elpoelma:
> @elpoelma It depends on the size of the agenda point contents, but yeah, if they aren't small, then we should only ask for the fields we need. Actually, now that I think of it, I'm maybe thinking in relational DB terms, where adding a field only costs as much as the time to read it and send it, but I guess in a triplestore it also adds lookup time, so we should always limit the fields we ask for if we don't need them all.

Yeah it would also limit the amount of data sent to the client... Of course, we would then require that the template `insert` method in the plugins package calls the `reload` method, which is maybe not ideal...

elpoelma:
> @elpoelma
> 
> > and the reload property was mutated to true due to recent code changes
> 
> I was not able to figure out why is this happening. Do you know?

Check-out this code segment: https://github.com/emberjs/data/blame/b176861834bc47da233b8fafe57d95a5c5d197fc/packages/model/src/-private/model-methods.ts#L60-L61

elpoelma:
> The fix at hand
> 
>     * Fixes the bug when creating reg statement
> 
>     * Does not degrade creating agenda point (because of the `fetchAll`)
> 
> 
> I do propose we go ahead with the fix, and I will create a ticket for a nicer solution?

I agree that we can go ahead with this solution, as a `reload` is not really needed for agendapoint templates. We can then later pick-up the ticket.

PR #614, opened 2024-01-05
Use undo history to track 'isDirty' instead of comparing html strings

piemonkey:
### Overview
Since we often add extra attributes to the html loaded into the editor, comparing the html input and that in the editor always fails, so we think the document has been modified. Since we clear the undo state on save, we can instead use this to determine whether all changes have been saved. This misses changes that have been manually undone, such as adding, then deleting the same text but this feels like an acceptable trade for the simplicity of this solution.

https://binnenland.atlassian.net/browse/GN-4016

### Setup
N/A

### How to test/reproduce
Open any agenda point or regulatory statement, optionally make some changes and undo them, then go back, you are no longer asked whether you want to leave with unsaved changes. If you have unsaved changes, it will ask to confirm.

### Challenges/uncertainties
As noted in the code, if we were to maintain the undo history, we would need to modify this functionality. Ideally this case would be supported directly by the editor.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/smooth-maps-type.md
+++ b/.changeset/smooth-maps-type.md
@@ -1,0 +1,5 @@
+---
+"frontend-gelinkt-notuleren": patch
+---
+
+Correctly track the saved state so we only ask to save changes when changes have been made
--- a/app/controllers/agendapoints/edit.js
+++ b/app/controllers/agendapoints/edit.js
@@ -94,6 +94,7 @@ import {
   GEMEENTE,
   OCMW,
 } from '../../utils/bestuurseenheid-classificatie-codes';
+import { undo } from '@lblod/ember-rdfa-editor/plugins/history';
 
 export default class AgendapointsEditController extends Controller {
   @service store;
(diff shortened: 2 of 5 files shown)


elpoelma (approved):
Very nice improvements, this should also (hopefully) solve the issues described in https://binnenland.atlassian.net/browse/GN-4169?atlOrigin=eyJpIjoiYjBkN2EwZWJlYzIzNGIyOTkyMWQ5NGM1NGQ0NjY4ZWUiLCJwIjoiaiJ9.

Could you just add the same logic to the `dirty` getter of the intro and outro pages of a meeting?

We should probably also include the same changes in a PR on the regulatory-statements frontend.

elpoelma:
Also solves [GN-4169](https://binnenland.atlassian.net/browse/GN-4169?atlOrigin=eyJpIjoiYjBkN2EwZWJlYzIzNGIyOTkyMWQ5NGM1NGQ0NjY4ZWUiLCJwIjoiaiJ9)

piemonkey:
I've updated the ones I missed and also created [a PR](https://github.com/lblod/frontend-reglementaire-bijlage/pull/214) on RS.

PR #621, opened 2024-01-22
Improve browser language detection and set nl-be fallback

elpoelma:
### Overview
This PR applies the same set of changes as https://github.com/lblod/ember-rdfa-editor/pull/1117, improving the detection of the user's browser language.

https://github.com/lblod/ember-rdfa-editor/pull/1117

### How to test/reproduce
Use [a locale switcher](https://addons.mozilla.org/en-US/firefox/addon/locale-switcher/) to try different combinations of locales and that it always translates the UI as expected. In that add-on you can select multiple locales by clicking the switch at the top:
image

### Checks PR readiness
- [x] Check if dummy app is correctly updated
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -1,6 +1,7 @@
 import Route from '@ember/routing/route';
 import { service } from '@ember/service';
 import ENV from 'frontend-gelinkt-notuleren/config/environment';
+import { decentLocaleMatch } from '../utils/intl';
 
 const featureFlagRegex = /^feature\[(.+)\]$/;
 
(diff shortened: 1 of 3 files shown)


app/routes/application.js
```diff
@@ -1,6 +1,7 @@
 import Route from '@ember/routing/route';
 import { service } from '@ember/service';
 import ENV from 'frontend-gelinkt-notuleren/config/environment';
+import { decentLocaleMatch } from '../utils/intl';
```
piemonkey, line 4: My intent was that we would import the function from the editor repo rather than having to duplicate it in every codebase. Is there a good reason not to do that?

piemonkey (commented):
Looks good, but not sure about importing vs duplication.

app/routes/application.js
elpoelma: Yeah I see, the main reason I did not import it from the editor is that the function is not really inherent to the editor codebase (it is quite generic).

PR #622, opened 2024-01-24
Add 'edit rdfa' mode, hard-coded to be switched off

piemonkey:
### Overview
Wire in the UI elements and fix some bits along the way.

There is a problem with some of the `besluit` nodes, as these need to be migrated to be rdfa aware. These will be in the plugins repo though.

Requires https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/379
In turn, requires https://github.com/lblod/ember-rdfa-editor/pull/1128

### Setup
Shouldn't need any as package.json has the git commits. It will take an age to `npm i` though. I think we should just dump this idea to be honest and just link...

### How to test/reproduce
All behaviour *should* be the same. There are some problems with `besluit` nodes though, which also has the side-effect of asking you each time, what decision type you want to use.

Then edit `{regulatory-statements,agendapoints}/edit.hbs` to switch `@shouldEditRdfa` to `true`. You should now see the new UI elements in the sidebar and be able to use them correctly.

### Challenges/uncertainties
Need to do more work in the plugins repo.

### Checks PR readiness
- [x] npm lint
- [x] no new deprecations


package.json
```diff
@@ -76,7 +77,7 @@
     "ember-drag-drop": "^0.9.0-beta.0",
     "ember-feature-flags": "^6.0.0",
     "ember-fetch": "^8.1.2",
-    "ember-intl": "^5.7.2",
+    "ember-intl": "^6.1.0",
```
elpoelma, line 80: There is an issue with the `ember-intl` 6.x packages at the moment. We should probably remain on 5.7.2 for now.

app/routes/regulatory-statements/edit.js
```diff
@@ -17,18 +17,18 @@ export default class RegulatoryStatementsEditRoute extends Route {
   }
 
   async model(params) {
-    const container = await this.store.findRecord(
+    const containerPromise = this.store.findRecord(
```
elpoelma, line 20: What is the reason for this change?

app/controllers/regulatory-statements/edit.js
```diff
@@ -1,9 +1,7 @@
 import Controller from '@ember/controller';
```
elpoelma, line 1: What is the reason for the import order change?

elpoelma (commented):
Looks good overall! (aside from some comments)
- Can you add changesets for the different parts you changed (deprecation resolving, promise-handling...)?

package.json
piemonkey: This was a confusion from the different version on the `feat/rdfa-links-ui` branch of the editor repo. I've corrected that now and I'll remove the commit from here.

app/routes/regulatory-statements/edit.js
piemonkey: When adding a debugger statement here to look into what was going on, I realised that instead of passing promises to RSVP, which will run them all in parallel, we were instead running them in series using `await`. Rather than just put them back the way they were I thought I'd improve loading times a little by changing it over.

app/controllers/regulatory-statements/edit.js
piemonkey: I was tired of the mess of imports, with newlines all over the place, so just snapped and arranged it by external/editor/plugins...

elpoelma: ok :smile:

piemonkey:
> Looks good overall! (aside from some comments)
> 
>     * Can you add changesets for the different parts you changed (deprecation resolving, promise-handling...)?

Sure, but these seem very internal changes. Who are the intended consumers of these changelogs? I figure on the editor, plugins and embeddable that we can assume we're talking to developers, but I'm not sure here.

app/routes/regulatory-statements/edit.js
elpoelma: Ok that makes sense!

elpoelma:
> > Looks good overall! (aside from some comments)
> > ```
> > * Can you add changesets for the different parts you changed (deprecation resolving, promise-handling...)?
> > ```
> 
> Sure, but these seem very internal changes. Who are the intended consumers of these changelogs? I figure on the editor, plugins and embeddable that we can assume we're talking to developers, but I'm not sure here.

Yeah I understand where you are coming from. The changelogs of our frontends/applications can of course also be a useful guide for us to quickly see what changed in each versions (this may include deprecation removals). This can also make it easier to track down/back issues.

@dkozickis @abeforgit what do are your opinions on this?

elpoelma (approved):

abeforgit:
In principle the rule is any change to master should have a changeset, so we can trace back indeed. I don't want to make the distinction internal vs external there cause the backtrace is very useful (and especially here everything is internal anyway since its a frontend). 
But a pr-on-pr like this I don't have strong opinions. As long as the eventual merge to master contains a good summary.

PR #625, opened 2024-02-04
Enable rdfa mode for rbs and some tweaks

abeforgit:
Mainly adding the snippet insert feature and the css import

### Setup

just npm i and run, the plugins are connected with github link

Note: to connect to the futurenow RB backend when running locally, you'll need to change the endpoints in config/environment.js
I debated adding that here but I wanna avoid frantic debugging due to forgetting this was changed.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/app/controllers/regulatory-statements/edit.js
+++ b/app/controllers/regulatory-statements/edit.js
@@ -58,6 +58,7 @@ import {
   templateComment,
   templateCommentView,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/template-comments-plugin';
+import { getActiveEditableNode } from '@lblod/ember-rdfa-editor/plugins/_private/editable-node';
 import {
   STRUCTURE_NODES,
   STRUCTURE_SPECS,
@@ -90,6 +91,7 @@ import {
   OCMW,
 } from '../../utils/bestuurseenheid-classificatie-codes';
 
+import SnippetInsertRdfaComponent from '@lblod/ember-rdfa-editor-lblod-plugins/components/snippet-plugin/snippet-insert-rdfa';
 export default class RegulatoryStatementsRoute extends Controller {
   @service documentService;
   @service store;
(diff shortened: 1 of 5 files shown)


app/controllers/regulatory-statements/edit.js
```diff
@@ -183,6 +186,12 @@ export default class RegulatoryStatementsRoute extends Controller {
     ];
   }
 
+  get activeNode() {
+    if (this.controller) {
+      return getActiveEditableNode(this.controller.activeEditorState);
```
piemonkey, line 191: `getActiveEditableNode()` seems to get `undefined` whenever an ember-node is selected, which then causes an exception in the snippet plugin, leaving the editor in a partially-working state. For example, by adding/clicking on a link node.

piemonkey (commented):
Creating a resource node and adding a `say:allowedSnippetList` named node property with snippet list uris from futurenow works. For some reason I'm seeing an intl warning for a translation that looks to exist in the plugins repo:
```
[ember-intl] translation: "snippet-plugin.alert.loading" on locale: "en-us, nl-be" was not found.
```

The crash noted in an inline comment is the only showstopper, though it's likely a fix in the plugins repo.

app/controllers/regulatory-statements/edit.js
elpoelma: getting the same error, will check if I can reproduce it in the plugins repo.

elpoelma: This should be fixed now. A conditional checking if `activeNode` is defined was missing.

elpoelma:
> Creating a resource node and adding a `say:allowedSnippetList` named node property with snippet list uris from futurenow works. For some reason I'm seeing an intl warning for a translation that looks to exist in the plugins repo:
> 
> ```
> [ember-intl] translation: "snippet-plugin.alert.loading" on locale: "en-us, nl-be" was not found.
> ```
> 
> The crash noted in an inline comment is the only showstopper, though it's likely a fix in the plugins repo.

Translation issue should also be fixed now.

elpoelma (approved):

PR #627, opened 2024-02-09
GN-4660: Percentage resize for tables instead of absolute pixels

dkozickis:

* https://binnenland.atlassian.net/browse/GN-4660
* https://binnenland.atlassian.net/browse/GN-4657

### Setup

1. Checkout
2. `npm install` 

### How to test/reproduce

Issues observed in below tickets should not happen anymore 

* https://binnenland.atlassian.net/browse/GN-4660
* https://binnenland.atlassian.net/browse/GN-4657

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/.changeset/cyan-rice-draw.md
+++ b/.changeset/cyan-rice-draw.md
@@ -1,0 +1,5 @@
+---
+"frontend-gelinkt-notuleren": patch
+---
+
+Update `@lblod/ember-rdfa-editor` to 9.2.0
--- a/.changeset/short-suits-fly.md
+++ b/.changeset/short-suits-fly.md
@@ -1,0 +1,5 @@
+---
+"frontend-gelinkt-notuleren": minor
+---
+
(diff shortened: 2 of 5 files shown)


elpoelma (approved):

elpoelma:
I don't really see the reason for changing `table-layout: fixed` to `table-layout: auto` though? I would think using `fixed` is better in this case as we specify the widths? In case of `auto` the browser can adapt them.

elpoelma:
> @elpoelma if last row has no set width and all other columns take 100% of width (which can happen if column is inserted but not resized) then the `fixed` logic will hide it from display completely :( using `auto` resolves that.
> CleanShot.2024-02-27.at.11.00.56.mp4

Ok that makes sense! Also noticed that we can achieve the same thing with `min-width: 100%` instead of `width: 100%`. I'm just fearing that setting the `table-layout` to `auto` will allow the browser to resize cells we do not want resized.

Issue #639, opened 2024-04-12
GN-4769: Error on signing extract


piemonkey:
When signing an extract the user is shown an error message even though the signing was successful:
'Error: Ember Data Request POST /signed-resources returned a 400 Payload (application/vnd.api+json; charset=utf-8) [object Object]'

User has to refresh the page manually to see the normal page.

### Reproduce
See https://binnenland.atlassian.net/browse/GN-4769

### Expected
When signing is successful > no error message is shown.

PR #640, opened 2024-04-12
Add support for signed/published-resource contents in files

piemonkey:
### Overview
Handle the possibility that published or signed resources have content in a file instead of directly in the resource as a string. Also added some very basic error handling for file lookups instead of just returning blank data.

Part of #639 
Needed for https://github.com/lblod/app-gelinkt-notuleren/pull/181

### Setup
Needs to be tested with the app-gn from the PR.

### How to test/reproduce
See [the app-gn PR](https://github.com/lblod/app-gelinkt-notuleren/pull/181) on how to test/reproduce.

### Challenges/uncertainties
The sheer number of deprecations makes it hard to see messages...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/routes/meetings/publish/publication-actions/detail.js
+++ b/app/routes/meetings/publish/publication-actions/detail.js
@@ -12,8 +12,13 @@ export default class MeetingsPublishPublicationActionsDetailRoute extends Route 
       },
     );
     const signedResource = await log.signedResource;
-    if (signedResource) {
+    if (signedResource?.content) {
       return signedResource.content;
+    } else if (signedResource?.file) {
+      const fileMeta = await signedResource.file;
+      const fileContent =
+        fileMeta && (await (await fetch(fileMeta.downloadLink)).text());
+      return fileContent;
     } else {
       const publishedResource = await log.publishedResource;
       return publishedResource.content;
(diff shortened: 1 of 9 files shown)


app/routes/meetings/publish/publication-actions/detail.js
```diff
@@ -12,8 +12,16 @@ export default class MeetingsPublishPublicationActionsDetailRoute extends Route
...
+      const fileMeta = await signedResource.file;
+      const fileReq = await fetch(fileMeta.downloadLink);
+      if (fileReq.ok) {
+        return fileReq.text();
+      } else {
+        return `<div class="au-c-alert au-c-alert--warning"><p>Error fetching file contents: ${fileReq.statusText}</p></div>`;
+      }
     } else {
       const publishedResource = await log.publishedResource;
       return publishedResource.content;
```
elpoelma, line 27: We probably also need to check for the `file` relationship here.

piemonkey: Good spot, yes.

app/routes/meetings/publish/publication-actions/detail.js
```diff
@@ -12,8 +12,16 @@ export default class MeetingsPublishPublicationActionsDetailRoute extends Route
       },
     );
     const signedResource = await log.signedResource;
-    if (signedResource) {
+    if (signedResource?.content) {
       return signedResource.content;
+    } else if (signedResource?.file) {
```
elpoelma, line 17: This condition will always be truthy as long as `signedResource` is defined, as the `file` relationship (event if its content is `null`) will always exist.

piemonkey: Good point. I've actually fixed this in my latest changes.

PR #641, opened 2024-04-12
GN-4772: Fix supported language detection

### LINKED ISSUE #642: GN-4772: Application is in English when it should be in Dutch
https://binnenland.atlassian.net/browse/GN-4772

dkozickis:
### Overview

User preferred order of languages (e.g. from `navigator.languages`) was not taken into regard if _more_ preferred language did not match supported language code exactly, but there is an exactly matching less preferred language.

E.g.  user `navigator.languages` returns `['nl-NL', 'en-US']` but supported languages are `['nl-BE', 'en-US']`, then `decentLocaleMatch` function would return `['en-us', 'nl-be']`, as `en-US` was matched and added to the matched languages first because of how the code was structured.

* https://binnenland.atlassian.net/browse/GN-4772
Closes #642 

### Setup

1. Checkout
2. `ember s --proxy https://dev.gelinkt-notuleren.lblod.info` 

### How to test/reproduce

You will need a Local Switcher extension installed. Set it up like so

CleanShot 2024-04-12 at 14 23 26@2x

Navigate to the application, it should be in Dutch. If you test without this PR then it will be in English. I've also added unit test to make sure this won't get broken again (hopefully).

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/.changeset/calm-llamas-protect.md
+++ b/.changeset/calm-llamas-protect.md
@@ -1,0 +1,5 @@
+---
+"frontend-gelinkt-notuleren": patch
+---
+
+GN-4772: Fix supported language detection
--- a/app/utils/intl.js
+++ b/app/utils/intl.js
@@ -1,34 +1,43 @@
+const matchUserLocaleToSupportedLocale = (userLocale, supportedLocales) => {
+  // Check if there is an exact match, e.g. nl-be === nl-be
+  const supportedLocale = supportedLocales.find(
+    (locale) => locale === userLocale,
+  );
+
+  if (supportedLocale) {
+    return supportedLocale;
+  }
+
+  // Check if there is a match based on language, e.g. nl === nl-be
+  const language = userLocale.split('-')[0];
+  return supportedLocales.find((locale) =>
+    locale.toLowerCase().startsWith(language.toLowerCase()),
+  );
+};
+
 /**
  * Helper function to help pick supported locales for ember-intl.
- * Returns exact matches first, then language matches, then the default.
+ *
+ * @param {string[]} userLocales - The user's locales, in order of preference,
+ * possibly from `navigator.languages`
+ * @param {string[]} supportedLocales - The locales the app supports, provided by `ember-intl`.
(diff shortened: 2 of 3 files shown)


piemonkey:
Hmm, I'm not sure that always taking the first approximate match is the correct thing here. It's a weird case, but technically the choice of language can be quite precise. For example, `['nl-NL', 'en-US', 'nl', 'en']` means 'give me NL Dutch, but if that precise match fails, give me American English, but if that fails, fall back to any Dutch or then English'. So I think the problem with the old implementation wasn't that it treated exact matches as higher priority than just matching the language, I think it's that it should treat a language-only locale (e.g. 'nl') as an exact match for anything with that language (e.g. nl-NL, nl-BE, etc).

So, I think (if we support 'en-US' and 'nl-BE'),
`['nl-NL', 'en-US', 'nl', 'en']` should give `['en-US', 'nl-BE']`,
`['nl-NL, 'nl', 'en-US', 'en']` should give `['nl-BE', 'en-US']`.
I think the first one is how the old code works, the second one it was broken but your code works.

Maybe this is just following (what I see as) the spec too closely though and your code is maybe more intuitive to users who don't necessarily care which flavour of a language they get.

abeforgit:
I think going out of spec here has benefits indeed... we do i18n as a bit of matter of principle, it's not even a requested feature from any users, and our users really are only government employees. I think the way you have it now is a fair compromise, always take the first fuzzy match. If we'd have a bunch of other languages, or if we'd actually take things like the difference between `nl-NL` and `nl-BE` seriously, it'd be worth doing it more proper, but we really don't. (And `nl-NL` vs `nl-BE` is not like `en-UK` vs `en-US`, there are afaik no spelling differences)

abeforgit (approved):

abeforgit:
You could even argue that _always_ showing dutch, with a special `en` route you can manually toggle, is a better ux for this website, as people with english browsers probably _still_ want to see the website in dutch

PR #669, opened 2024-05-30
Correctly tag release builds as latest in docker hub

piemonkey:
### Overview
The release woodpecker build is only pushing to the specific version tag, not to `:latest`, so `:latest` is actually pushed from the latest script. Woodpecker doesn't seem to run latest for the release commit, presumably either because it triggers the release pipeline, or because it's pushed at the same time, so there's only one 'push' event (not sure why it wouldn't pick the latest pushed commit though).
So until the first change to master after a release, `:latest` will always misreport as being the previous version.

N/A

### Setup
N/A

### How to test/reproduce
Push a new tag and verify in docker hub that the shas for both that version and `latest` are the same.

### Challenges/uncertainties
I don't know for sure if this will work, their docs don't say how to specify multiple tags, but their code looks like just comma seperating works.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.changeset/long-ligers-argue.md
+++ b/.changeset/long-ligers-argue.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Correctly tag release builds as latest
--- a/.woodpecker/.release.yml
+++ b/.woodpecker/.release.yml
@@ -3,7 +3,7 @@ steps:
     image: woodpeckerci/plugin-docker-buildx
     settings:
       repo: ${CI_REPO}
-      tags: "${CI_COMMIT_TAG##v}"
+      tags: [ "latest", "${CI_COMMIT_TAG##v}" ]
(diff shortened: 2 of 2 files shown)


piemonkey:
> As far as I can see they just add tags to `docker build` command line
> 
> [codeberg.org/woodpecker-plugins/docker-buildx/src/branch/main/plugin/docker.go#L142](https://codeberg.org/woodpecker-plugins/docker-buildx/src/branch/main/plugin/docker.go#L142)
> 
> so `-t one_tag -t two_tag`, which is the way to go with `docke build`

Yes, but I can't see whether `build.Tags.Value()` in that code is going to return an array of tags or just a single tag which contains a comma.

Now I realise that I'm missing the obvious. The format is yml, which is essentially a badly designed programming language, so I can just specify an array. I'll change it.

PR #677, opened 2024-06-11
Switch to location-plugin with map for address variables

piemonkey:
### Overview
Uses the new nodespec and components from the location plugin for addresses. Any address variables from templates get correctly picked up by the plugin and treated as locations instead.

Includes an update to ember-power-select as the old version caused a mismatch in a mutual dependency.

https://binnenland.atlassian.net/browse/GN-4840

### Setup
N/A

### How to test/reproduce
Create/edit an agenda point or regulatory statement, insert locations, edit them, jump between 'places' and 'addresses', etc. Create a regulatory statement from a template including an address variable to confirm that it's still functional as a location.
Keep an eye out for any problems with power-select driven components.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


package.json
```diff
@@ -46,7 +47,7 @@
     "@lblod/ember-environment-banner": "^0.5.0",
     "@lblod/ember-mock-login": "0.7.0",
     "@lblod/ember-rdfa-editor": "9.10.0",
-    "@lblod/ember-rdfa-editor-lblod-plugins": "19.2.0",
+    "@lblod/ember-rdfa-editor-lblod-plugins": "19.2.0-dev.7de1e6b44befd7da2b55137b1576c179b4624e9a",
```
lagartoverde, line 50: This shouldn't be merged until the plugins are released then right?

PR #679, opened 2024-06-14
Update to ember-modifier 4.1 and fix some deprecations

piemonkey:
### Overview
Doesn't fix all the deprecations from ember-data or even the one from AuLoader but `async: true` was the worst culprit and could be done completely brainlessly. The need to change ember-modifier peer-deps was the only part that took time.
Draft until a new editor release happens.

N/A

### Setup
N/A

### How to test/reproduce
This should disable at least most of the other ember-data deprecations to give you the zen developer experience you've always wanted (and to check that I've got rid of the async and modifier ones) in app/app.js:
```
import { registerDeprecationHandler } from '@ember/debug';

registerDeprecationHandler((message, options, next) => {
  if (
    [
      'The `firstObject` method on the class ManyArray is deprecated. Use the native array method `[0]` instead.',
      'The `firstObject` method on the class RecordArray is deprecated. Use the native array method `[0]` instead.',
      'The `toArray` method on the class RecordArray is deprecated. Use the native array method `slice` instead.',
      'The `toArray` method on the class ManyArray is deprecated. Use the native array method `slice` instead.',
      'The `sortBy` method on the class RecordArray is deprecated. Use the native array method `.slice().sort` instead.',
      'The `sortBy` method on the class ManyArray is deprecated. Use the native array method `.slice().sort` instead.',
      'Do not use A() on an EmberData PromiseManyArray',
      'The `pushObject` method on the class ManyArray is deprecated. Use the native array method `push` instead.',
      'The `pushObjects` method on the class ManyArray is deprecated. Use the native array method `push` instead.',
      "The pushObject method on ember-data's PromiseManyArray is deprecated. await the promise and work with the ManyArray directly.",
      "The map method on ember-data's PromiseManyArray is deprecated. await the promise and work with the ManyArray directly.",
      'The `setObjects` method on the class ManyArray is deprecated. Use the native array method ``arr.length = 0; arr.push(objects);`` instead.',
      'The `clear` method on the class ManyArray is deprecated. Use the native array method `length = 0` instead.',
      "The toArray method on ember-data's PromiseManyArray is deprecated. await the promise and work with the ManyArray directly.",
    ].includes(message) ||
    message.startsWith(
      '[AuLoader] This version of the `AuLoader` component is deprecated.',
    )
  ) {
    return;
  }
  console.log(message);
  return next(message, options);
});
```

### Challenges/uncertainties
The further deprecation fixes won't be hard, just require some second thought as to whether they're doing the right thing.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


lagartoverde (approved):
Looks good to me

lagartoverde:
Maybe we should add a ticket on Jira to remove the remaining deprecations? @abeforgit

elpoelma (changes requested):
Looks good overall.
A small remark on the usage of `ember-modifier` 4.x: it seems that the types/migration-guide of `ember-modifier` 4.x are/is not entirely correct (the `{ eager: false; }` options seems no longer supported). It seems to me that it is not trivial to support both `ember-modifier` in the editor/editor-plugins packages, as the types will fail when `ember-modifier` 4.x is installed. This is however not an issue for this application, as it doesn't use typescript. We should probably look into using https://www.npmjs.com/package/@embroider/macros for the editor and editor-plugins to support both the 3.x and 4.x range OR we should look into dropping the 3.x range.

piemonkey:
@elpoelma I think the `eager: false` thing is more confusing than it seems. [It looks like](https://github.com/ember-modifier/ember-modifier/pull/222) rather than `eager: false` not being supported, `eager: false` is the new default, so since we're currently using that everywhere, there should be no changes for us. At runtime, if any code is passing the options, they will just be ignored as an extra argument.

For the types, if I switch to `4.1.0` on the plugins test app and remove `{ eager: false }`, then the type checks pass. This is expected, as the type checks should not be type checking the dependencies. Are there any problems that you're seeing that I'm missing?

elpoelma:
> @elpoelma I think the `eager: false` thing is more confusing than it seems. [It looks like](https://github.com/ember-modifier/ember-modifier/pull/222) rather than `eager: false` not being supported, `eager: false` is the new default, so since we're currently using that everywhere, there should be no changes for us. At runtime, if any code is passing the options, they will just be ignored as an extra argument.
> 
> For the types, if I switch to `4.1.0` on the plugins test app and remove `{ eager: false }`, then the type checks pass. This is expected, as the type checks should not be type checking the dependencies. Are there any problems that you're seeing that I'm missing?

I guess it's not really a problem in this case, as `skipLibCheck` is set to `true`, I guess I was mostly assuming the hypothetical where `skipLibCheck` is set to `false`. (in case of other typescript projects using the editor package)

PR #681, opened 2024-06-18
Fixed document service not targeting the correct predicates

lagartoverde:
### Overview
With the change of the editor, we stopped using relative uris like `a` and started using `'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'` instead. We already make this fix on the plugins repo but failed to do here. We might need to do it somewhere else but don't know where so if someone knows another place is targeting relative uris let me know.

GN-4864

### Setup
None

### How to test/reproduce
Create an agendapoint, write a description on it, then go to a meeting and try to add the agendapoint and copy its description. It should be copied correctly while previously it was not

### Challenges/uncertainties
As I said we could be missing the same fix on other places 

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/twelve-snakes-joke.md
+++ b/.changeset/twelve-snakes-joke.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Fixed copy agendapunt description button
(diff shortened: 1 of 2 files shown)


elpoelma (approved):
Looks good to me! I'm a bit surprised that the triple-extractor does not expand `a` to `http://www.w3.org/1999/02/22-rdf-syntax-ns#type` automatically.

PR #635, opened 2024-03-28
GN-4748: basic support for fetching dynamic decision templates

elpoelma:
### Overview
This PR adds basic support for fetching dynamic decision templates from the regulatory statement endpoint.
Specifically it includes the following changes/additions:

**A refactor of the `RegulatoryStatementsFetcher` service**
- Renamed to `TemplateFetcher` to better reflect its purpose
- Add support for fetching templates of different types (not only regulatory statement templates)
- Add support for querying the `contexts`/`disabledInContexts` properties of templates (important for decisions/agendapoints)

**Modifications to models of agendapoint routes**
- Modified `new` route of agendapoints to also fetch decision templates from regulatory statement endpoint
- Modified `edit` route of agendapoints to also fetch decision templates from regulatory statement endpoint

I made the following initial assumptions in this PR:
- The type of a decision template is `http://data.lblod.info/vocabularies/gelinktnotuleren/BesluitTemplate`
- The predicates used to specify template contexts are: `ext:context` and `ext:disabledInContext`

@abeforgit section:

** Modifications to support new structure node, and the removal of the decision nodes in version 22 of the plugins
** bumps and config changes to support v10 editor and v22 plugins
- merge of @piemonkey 's PR to get the deps to be friends
- changed the builder image cause our ember images don't have a node v20 version

##### connected issues and PRs:
[GN-4906](https://binnenland.atlassian.net/browse/GN-4906)
[GN-4748](https://binnenland.atlassian.net/browse/GN-4748?atlOrigin=eyJpIjoiMWM0ZjhiYzU0ZjZmNDYyYmIwNGIzZDM4NDM5YzdlNWIiLCJwIjoiaiJ9)
https://github.com/lblod/frontend-reglementaire-bijlage/pull/236
https://github.com/lblod/app-reglementaire-bijlage/pull/77
https://github.com/lblod/reglement-publish-service/pull/14
Solves https://github.com/lblod/frontend-reglementaire-bijlage/issues/227

### Setup
Check-out https://github.com/lblod/app-reglementaire-bijlage/pull/77 for a description on how to set-up the stack

### How to test/reproduce
Check-out https://github.com/lblod/app-reglementaire-bijlage/pull/77 for a description on how to test the new features.

### Checks PR readiness
- [x] UI: works on smaller screen sizes
- [x] UI: feedback for any loading/error states
- [x] Check cancel/go-back flows
- [x] Check database state correct when deleting/updating (especially regarding relationships)
- [x] changelog
- [x] npm lint
[...]

PR #683, opened 2024-06-26
Set-up basic creation-flow of installation meetings for municipalities

elpoelma:
### Overview
This PR sets-up a basic creation-flow for installation meetings concerning municipal councils (gemeenteraden).
Specifically, it adds the following:
- Updates to the ember-data model definitions in order to support 'installatievergadering' resources.
- Modifications to the meeting inbox page in order to correctly show installation meetings and support the creation of installation meetings alongside common meetings.
- Support the creation of a basic installation meeting concerning a municipal council. At the moment, these meetings contain dummy hard-coded agendapoints.
- Some small adjustments to the meeting page when an installation-meeting is opened (mostly in terms of vocabulary used)

TODOS:
- [x] Internationalization
- [ ] (In the future) Move away from hard-coded meetings
- [ ] (In the future) Support different types of installation meetings.

[GN-4828](https://binnenland.atlassian.net/browse/GN-4828)
https://github.com/lblod/app-gelinkt-notuleren/pull/187

### Setup
This PR should be tested in combination with https://github.com/lblod/app-gelinkt-notuleren/pull/187

### How to test/reproduce
- Set-up the backend and start the frontend
- Log in as a municipality and visit the meeting inbox page
- Using the new meeting-creation menu, create a sample installation meeting
- Ensure that an installation meeting has been created with 9 (empty) sample agendapoints. 
- Notice that some terms have been adapted on the meeting-edit page to reflect you are editing an installation meeting
- On the meeting inbox page, you should be able to differentiate common meeting and installation meetings.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


app/components/agenda-manager/edit.js
```diff
@@ -19,6 +19,7 @@ export default class AgendaManagerEditComponent extends Component {
   }
 
   submitTask = task(async (item) => {
+    console.log('Item: ', item);
```
piemonkey, line 22: Console.log

app/controllers/inbox/meetings.js
```diff
@@ -14,4 +16,60 @@ export default class InboxMeetingsController extends Controller {
   get readOnly() {
     return !this.currentSession.canWrite && this.currentSession.canRead;
   }
+
+  isInstallationMeeting = (meeting) => {
+    return meeting instanceof InstallatieVergaderingModel;
+  };
+
+  @action
+  async createInstallationMeeting() {
```
piemonkey, line 25: This method takes a while, even locally. I think we need to add something to give feedback to the user that something is happening. Only part of the delay is down to this method, as it also looks like there's something of a waterfall of requests after it has created the meeting. Strangely it doesn't have such a strong impact when simply loading an existing meeting.

translations/en-us.yaml
```diff
@@ -141,18 +141,22 @@ inbox:
     time: Time
     location: Location
     meeting-of: Meeting of
+    installation-meeting: Installation meeting
```
piemonkey, line 144: I also heard 'Inauguration meeting', which seems to fit better as a translation for me. I could be wrong on the exact dutch meaning though.

piemonkey (commented):
Looks good, I think it just needs some feedback to the user on loading.

app/controllers/inbox/meetings.js
piemonkey: Actually, looking more closely, it's the POST requests to create the new agendapoints and treatments that seem to be cascading. It would be nice if we can get ember-data to handle this better, but that's likely a slippery slope...

translations/en-us.yaml
elpoelma: Good point, I updated it.

app/controllers/inbox/meetings.js
elpoelma: Yeah... I tried to batch the requests as much as possible (using `Promise.all`), but do not really know if there is a better way.
Will see if I can add a loading indicator.

abeforgit (approved):
tried this out, also in conjunction with #635 and it works nicely

elpoelma:
> tried this out, also in conjunction with #635 and it works nicely

Alright :) Should we hide this functionality using a feature flag?

abeforgit:
> > tried this out, also in conjunction with #635 and it works nicely
> 
> Alright :) Should we hide this functionality using a feature flag?

no I don't think we need to

abeforgit:
woj I was pushing my stuff to the wrong branch :man_facepalming:

PR #689, opened 2024-07-18
Move to static file service

lagartoverde:
### Overview
Moved the base image of the frontend to be the static-file-service based on some discussions on mu-semtech. 

None

### Setup
Build the image and serve it with the backend, now we can put the identifier as the main point of entry of the app

### Challenges/uncertainties
This would mean the user can access the different endpoints we use for data, not the most important thing, but something to take into account, I think we do the same with other apps

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (approved):
Works as expected in combination with the backend PR.

PR #700, opened 2024-08-16
Don't make unnecessary history changes when updating RS inside agendapoints

lagartoverde:
…

### Overview
When checking for RS updates we were making unnecesary changes and adding them to the history, which resulted in the editor thinking there were changes when there were not. This interfiered with the saving action.
I limited these changes, and I added a way of updating properties without adding them to the history

GN-4961
https://github.com/lblod/ember-rdfa-editor/pull/1214

### Setup
Link the ember-rdfa-editor with the branch of the PR I linked

### How to test/reproduce
Go to the agendapoints, create a new agendapoint with a RS, save it and check that it doesn't say that it has unsaved changes anymore, modify the RS go back to the agendapoint and check that now you have a unsaved change to save

### Challenges/uncertainties
Maybe in the future we shouldn't require user action in order to update the RS

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/components/editor-plugins/regulatory-statements/view.js
+++ b/app/components/editor-plugins/regulatory-statements/view.js
@@ -29,8 +29,10 @@ export default class ReadOnlyContentSectionComponent extends Component {
     if (this.node.attrs.title !== this.currentVersion.title) {
       this.args.updateAttribute('title', this.currentVersion.title);
     }
-    if (this.content !== this.currentVersion.htmlSafeContent) {
+    if (this.node.attrs.oldContent !== this.currentVersion.htmlSafeContent.toString()) {
       this.args.updateAttribute('content', this.currentVersion.htmlSafeContent);
+    } else {
+      this.args.updateAttribute('content', this.currentVersion.htmlSafeContent, true);
     }
   }
   get node() {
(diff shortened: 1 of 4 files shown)


app/components/editor-plugins/regulatory-statements/view.js
```diff
@@ -29,8 +29,17 @@ export default class ReadOnlyContentSectionComponent extends Component {
...
+      this.node.attrs.oldContent !==
+      this.currentVersion.htmlSafeContent.toString()
+    ) {
       this.args.updateAttribute('content', this.currentVersion.htmlSafeContent);
+    } else {
+      this.args.updateAttribute(
+        'content',
+        this.currentVersion.htmlSafeContent,
+        true,
+      );
```
elpoelma, line 42: Shouldn't these two be switched around?

elpoelma (changes requested):
With the if statement switched around, this seems to work as expected.
Some ideas/comments on another approach we could take:
- When inserting a regulatory statement in a decision, directly populate the `content` attribute (without doing it using the `did-insert` hook
- When saving a regulatory statement, we *could* update the all the linked documents in a headless way => this gives us some advantages:
  * No longer required to open agendapoint to update linked RS
  * We are sure AP's contain the latest version of a RS
  => Note: we should only do this with unpublished/unsigned AP's

elpoelma:
Ok, so essentially we could alter the code so it only updates the 'content' attribute when the content has actually changed.

elpoelma (approved):
Works, this PR no longer needs the editor bump then right?

PR #719, opened 2024-09-23
dev-tooling: add support for `dotenv` when in environment.js file

elpoelma:
### Overview
Add support for loading environment variables from a `.env` file in `environment.js`. A `.env` file is only loaded when `environment` is 'development'.

The following variables are supported:
- `MOW_REGISTRY_SPARQL_ENDPOINT`
- `PUBLICATION_SPARQL_ENDPOINT`
- `ROADSIGN_IMAGE_BASE_URL`
- `FALLBACK_CODELIST_SPARQL_ENDPOINT`
- `REGULATORY_STATEMENT_ENDPOINT`

This ensures that a developer can easily override some variables inside a `.env` file (which is part of `.gitignore`), without needing to adjust the `environment.js` (and accidentally committing it).

None

### How to test/reproduce
- Try to add some of the above variables to your `.env` file and ensure the application uses them correctly.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -1,5 +1,4 @@
 'use strict';
-
 module.exports = function (environment) {
   const ENV = {
     modulePrefix: 'frontend-gelinkt-notuleren',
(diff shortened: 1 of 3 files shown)


lagartoverde (approved):

elpoelma:
> Do we need changeset for this? debating myself if it's needed with an internal tag, or just shouldn't include it at all

I don't really think it is necessary, as it is really an internal tool, and not a big/drastic/breaking change either.

PR #722, opened 2024-10-02
Feat/download meeting part

lagartoverde:
### Overview
Added a way of downloading different parts of the meeting as HTML
https://www.figma.com/design/UVPMVe7O29nt7G1rRJ6uH9/installatievergadering?node-id=351-337&t=M0DRZAdacQNR1WNw-4
GN-4902

### Setup
None

### How to test/reproduce
Go to a meeting, click on the acties menu, and then on the "Onderdelen kopiëren en downloaden" you should be able to download all the meeting parts there

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
3 more things:
- I've merged @piemonkey 's work on the ability to delete template comments from the published html. You should use this for the decision export, for all the others the prepublisher already handles this
- again for decisions only: somewhere in the publication pipeline we replace the decision URI from "http://example.net/new-decision" with a real, new decision URI. We should do this logic for the decision download as well.
- and also for decisions: we should wrap it in a div that defines the prefixes we use. again, for all the other things this is something we already do somewhere else in the app

abeforgit (changes requested):
another thing: the agenda items are not in order on the export screen

PR #728, opened 2024-10-09
Add meetings/download/copy route with just the 'copy all' button

piemonkey:
### Overview
Do not split up the document parts yet as this way the copy-all functionality and the presence of the page can be demoed. Makes use of the download-meeting-part component, adjusted to allow for putting on the clipboard instead of only downloading. Also includes some deprecation fixes that I added...

Part of https://binnenland.atlassian.net/browse/GN-5067

### Setup
N/A

### How to test/reproduce
Go to the export options for a meeting page. All existing (just added) functionality should work, but the 'more copy options' button should take you to the new screen. After clicking 'copy full decision' the content should be on your system clipboard.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


translations/nl-BE.yaml
```diff
@@ -801,7 +798,15 @@ download:
...
     download-html: Download HTML
     no-data-message: Geen agendapunten gevonden in zitting
+    copy-options: Bekijk kopieermogelijkehden
 
 download-meeting-part:
   downloading: Aan het downloaden
   downloaded: Gedownload
+
+copy-options:
+  heading: 'Kopieer onderdelen uit: {besluit}'
```
abeforgit, line 808: the dutch version has a format param, the english one doesn't, which throws when you switch to dutch

abeforgit (changes requested):
- It's arguable whether the user would want to include the template comments or not, but I think in the way the usecase is currently envisioned we should strip them out

- image
some color mismatch here (I know its nitpicking but it's likely to be the first comment from the pms :grimacing: )
- also in the above: both columns have the same title

translations/nl-BE.yaml
abeforgit: worded that a bit weird, but its just that it expects a param in dutch which you're not giving

piemonkey: Oh, I forgot to remove the format param from the dutch when I decided to highlight the title in the same way as the heading of the download page, so had to move it out.

piemonkey:
> * It's arguable whether the user would want to include the template comments or not, but I think in the way the usecase is currently envisioned we should strip them out

Huh, it's using the same logic as the download page, so I never actually checked if it was stripping the comments. I'll change it (which will change it for both).

>       some color mismatch here (I know its nitpicking but it's likely to be the first comment from the pms 😬 )

Huh, good spot. It turns out giving an invalid `@skin` value leaves `<AuLInk>` to use default link styling. If only we had typescript...

>     * also in the above: both columns have the same title

Oops.

translations/en-us.yaml
```diff
@@ -797,7 +794,16 @@ download:
...
 download-meeting-part:
   downloading: Downloading
   downloaded: Downloaded
+
+copy-options:
+  heading: 'Copy parts from:'
+  disclaimer: 'The formatting of the document will depend on the application in which it is loaded.'
+  copy-all: Copy full decision
+  copying: copying
+  completed: copied
```
elpoelma, line 809: The translation for 'completed' is missing in 'nl-BE'

elpoelma (changes requested):
Some minor comments regarding translations etc.

I did notice that the template comments are still included when copying/download a decision.

piemonkey:
I'll fix the translation issues. The comment removal is in 2 parts. One is that I forgot to include the removal of comments from the preview, I'll fix that. In the copy/download though, I suspect the problem you're seeing is that due to the approach we took of marking the element for removal, the document must be created from a version of the template saved after the code change, otherwise it lacks the attribute to mark it for removal. Can you confirm whether the template you're looking at has this?

translations/en-us.yaml
piemonkey: Oops, I substituted 'completed' for 'copied'...

app/routes/meetings/download.js
```diff
@@ -5,9 +5,10 @@ export default class MeetingsDownloadRoute extends Route {
   @service store;
   async model(params) {
     const zitting = await this.store.findRecord('zitting', params.id, {
-      include: 'bestuursorgaan,agendapunten.behandeling',
+      include:
+        'bestuursorgaan,agendapunten.behandeling,agendapunten.behandeling.document-container',
     });
-    zitting.agendapunten = (await zitting.agendapunten).sortBy('position');
+    zitting.agendapunten = [...(await zitting.agendapunten)].sortBy('position');
```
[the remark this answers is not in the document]
piemonkey: True

translations/nl-BE.yaml
```diff
@@ -801,7 +798,16 @@ download:
     agendapoint-title: Titel agendapunt
     download-html: Download HTML
     no-data-message: Geen agendapunten gevonden in zitting
+    copy-options-header: Kopieer naar klembord
+    copy-options: Bekijk kopieermogelijkehden
```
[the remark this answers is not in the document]
piemonkey: Looks like a typo in the design, fixed.

piemonkey:
I fixed the preview issue now. I've tested it with a locally published template from RB master and it removes the comment correctly.

PR #732, opened 2024-10-15
Feat/manual voting

lagartoverde:
### Overview
Added a way to add a custom voting to the behandeling that will be just a document

https://binnenland.atlassian.net/browse/GN-4991
https://github.com/lblod/app-gelinkt-notuleren/pull/198

### Setup
You should use the https://github.com/lblod/app-gelinkt-notuleren/pull/198 as backend

### How to test/reproduce
Go to the meeting form, and notice that you should be able to add custom votings that just let you write some text in the a document

### Challenges/uncertainties
We should maybe add some extra rdfa in the document, I will have a look when I finish the prepublisher and see how the final document looks like
As I said the prepublisher is not finished, so that's something I have to finish before merging this PR

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
**bug**:
- add custom voting to an ap which doesn't have participants configured
- voting doesn't get added, but participant button is now greyed out
**bug**:
- you can now add a normal voting even if there are no participants, which was previously not allowed (and also doesn't work when you try to save the voting)

**bug**:

image
my browser is in dutch, but "custom voting" is not translated
**bug**:
image
pill not aligned nicely
**bug**:
image
in general could use a bit of padding or something to make it clear this is an input field

**bug**:
- edit title, and directly click on meeting "save" button
- "saved" pill shows, but title is not changed


**bug**:
"save" button does nothing, only "save and quit" works

**bug**:
not sure what's going on here:
image

lagartoverde:
Fixed all bugs and made a custom view for published votings

abeforgit (changes requested):
errors in the console when you add a new voting...
image

abeforgit (changes requested):
saving title is still broken:
repro 1:


- reset cache to get a consistent state
- change title, don't click the checkmark
- click the save (not the save&quit) button
- exit the document
- notice title is not changed
- page refresh and cache restart don't help


repro 2:
- change title
- don't click checkmark
- now click "save&quit"
- notice title still hasn't changed in overview
- refresh: now title is changed in overview
- BUT: open document: notice title is not changed there
- restart cache
- now title is changed in both places

abeforgit (changes requested):
bug:

- make 2 custom votings with some different content
- open 1 of them, click "Save" (not save & quit)
- open the other -> it's showing the wrong document

resolves itself when you use save&quit or refresh the page

abeforgit:
added a commit to fix the person plugin after I merged master, as we no longer use the vendor-proxy (it was also broken before though, as it was missing the nodespec in the config :grimacing: )

lagartoverde:
Fixed that bug, hopefully is the last one

PR #735, opened 2024-10-18
Add focus mode to the meeting form

lagartoverde:
### Overview
Add focus besluit mode to the meeting form

GN-5091

### Setup
None

### How to test/reproduce
Go to a meeting, you will have a switch on top, verify that when you activate the switch it correcly hides everything except the steps 1 3 and 5, also the publish and the copy buttons should be switched

### Challenges/uncertainties
If we expect more custom modes to be added we could create a variable to show and hide each block and with the focus mode hide the sections we want, I didn't do it because I think there are no plans on the horizon for this, but I could be wrong.
Also awaiting for confirmation on the text of the tooltip

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
the toggle state doesn't persist...

PR #737, opened 2024-10-21
Fixes to copy parts functionality

piemonkey:
### Overview
There are 3 fixes here of increasing subtlety:
1. ~~Set the copy-all button on the copy screen to set the plaintext mime type as well as text/html.~~
2. Prevent articles from before the `data-say-structure-content` attribute was added to article nodes from crashing the component.
3. Extract the `outerHTML` of the section before extracting parts, as `append()`ing of content elements to a `div` element in the callback function removes them from the element, and so from the `outerHTML` if we generate it after doing this.
4. Working around Firefox stripping many RDFa attributes when using the clipboard API.

Fixes the result of https://github.com/lblod/frontend-gelinkt-notuleren/pull/731
Fixes: https://binnenland.atlassian.net/browse/GN-5127

### Setup
N/A

### How to test/reproduce
For any meeting, go to the 'copy parts' screen for an agendapoint with a correct motivation section. How to test the two relevant fixes:
1. ~~Click the 'copy full decision' button and look at the `text/plain` clipboard contents~~
3. copy the whole motivation section (not the individual parts). Previously this was only extracting the headings, now it is the whole content.
4. Verify that on Firefox attributes such as `property` are maintained in the output.
To test 2. open the same screen for an agendapoint created more than 4 months ago, and note that the screen works. Previously the content was blank and there was an error in the console.

### Challenges/uncertainties
It's really unclear from the docs why 3. is actually happening, as it's not clear that appending a node to an element removes it from its previous parent.
The clipboard API on Firefox strips attributes in a way that doesn't seem to be documented anywhere, or at least, not anywhere that I could find.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/controllers/meetings/download/copy.js
+++ b/app/controllers/meetings/download/copy.js
@@ -7,7 +7,7 @@ import { copyStringToClipboard } from '../../../utils/copy-string-to-clipboard';
 export default class MeetingsDownloadCopyController extends Controller {
   @action
   async copyToClipboard(text) {
-    await copyStringToClipboard({ html: text }, true);
+    await copyStringToClipboard({ html: text });
   }
 
   get previewDocument() {
--- a/app/controllers/meetings/download/copy.js
+++ b/app/controllers/meetings/download/copy.js
@@ -7,7 +7,7 @@ import { copyStringToClipboard } from '../../../utils/copy-string-to-clipboard';
 export default class MeetingsDownloadCopyController extends Controller {
   @action
   async copyToClipboard(text) {
-    await copyStringToClipboard({ html: text });
+    await copyStringToClipboard({ html: text }, true);
   }
 
   get previewDocument() {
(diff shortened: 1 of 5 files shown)


app/controllers/meetings/download/copy.js
```diff
@@ -7,7 +7,7 @@ import { copyStringToClipboard } from '../../../utils/copy-string-to-clipboard';
 export default class MeetingsDownloadCopyController extends Controller {
   @action
   async copyToClipboard(text) {
-    await copyStringToClipboard({ html: text });
+    await copyStringToClipboard({ html: text }, true);
```
abeforgit, line 10: this function only takes a single object argument, unless some git shenanigans happened and I'm not seeing the changes you made to it

you maybe meant to add `renderHtmlToPlain: true`? but that's also true by default, so I'm not sure this is really doing anything

abeforgit (changes requested):

app/controllers/meetings/download/copy.js
piemonkey: Oof, I must have been tired when I put this in...

abeforgit (approved):

PR #740, opened 2024-10-23
fix(lmb-tables): adjust fractie-splitsing query

elpoelma:
### Overview
This PR introduces some bugfixes for the 'fractie-splitsing' LMB table query:
- Adjustment in how verkiezing and kieslijsten are determined (previous approach was erroneous)
- Fix an incorrect variable name

None

### How to test/reproduce
- Start the GN application
- Create/open an inauguration meeting
- Insert an LMB table with the `IVGR5-LMB-1-splitsing-fracties` tag
- When syncing, ensure the following
  * The correct election resource is used to determine the kieslijsten + fracties
  * It runs without error

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/clean-kings-turn.md
+++ b/.changeset/clean-kings-turn.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Implement a more extendable way to determine the correct bestuursorgaan-classificatie for a new inauguration meeting
--- a/.changeset/slow-seals-bathe.md
+++ b/.changeset/slow-seals-bathe.md
@@ -1,0 +1,7 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+**IVGR LMB tables: adjust fractie-splitsing query**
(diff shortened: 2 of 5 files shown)


abeforgit:
@elpoelma could you give a quick rundown of how the previous query behaved incorrectly?
Do I understand it correctly that we previously incorrectly assumed the election would have a single constant URI for all municipalities, but in fact they will all have their own URI?

elpoelma:
> @elpoelma could you give a quick rundown of how the previous query behaved incorrectly? Do I understand it correctly that we previously incorrectly assumed the election would have a single constant URI for all municipalities, but in fact they will all have their own URI?

Yes that is correct. I indeed incorrectly assumed there was only a single election URI. But there is a seperate election URI for each bestuursorgaan.
So:
- Based on the logged in user, we need to determine from which bestuursorgaan and connected election resource we need to fetch the kieslijsten.

abeforgit:
makes sense

abeforgit (approved):

PR #784, opened 2024-12-03
Sort IV mandatee tables by fractie votes when having equal seat numbers

piemonkey:
### Overview
Add sorting based on the sum of the votes for the candidates within a fractie to any mandatee table query where the fracties are sorted by number of seats. I think I haven't missed any that should be sorted this way, but it's possible that I missed some.

Notes:
- Some of the queries are significantly slower now. If we think this is a problem, we'll need to look to optimise the queries. Alternatively, it doesn't make sense to me to overload the queries in this way, it would make sense to do one query for the sorting of the fracties and then handle sorting based on this in code.
- There was one RMW query which was using the gemeenteraadslid functie code. I assumed this was a mistake, but I'll flag it in the review so it can be sanity checked.

Jira ticket: https://binnenland.atlassian.net/browse/GN-5308

### Setup
N/A

### How to test/reproduce
Sync mandatee tables for an IV and see that the new sorting is applied. For example, for Mechelen GR the previous sorting put PVDA above cd&v, but the new sorting puts cd&v first as they got more total votes.

### Challenges/uncertainties
It's hard to know if the numbers I'm seeing for votes are correct as I have no way to verify them. So, it's possible that I'm including results that I shouldn't or excluding ones that I should. I used similar logic to where we are getting vote numbers elsewhere, so I think it's unlikely, but it's hard for me to exclude it.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/config/mandatee-table-config.js
+++ b/app/config/mandatee-table-config.js
@@ -25,6 +25,7 @@ import {
   BESTUURSEENHEID_CLASSIFICATIE_CODES,
 } from './constants';
 import { promiseProperties } from '../utils/promises';
+import { fractieOrderingSubquery } from './mandatee-table-query-fragments';
 
 export const IVGR_TAGS = /** @type {const} */ ([
   'IVGR2-LMB-1-geloofsbrieven',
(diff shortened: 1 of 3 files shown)


app/config/mandatee-table-config.js
```diff
@@ -1670,20 +1633,12 @@ export const mandateeTableConfigRMW = (meeting) => {
             }
             ?bestuursorgaan besluit:bestuurt <${bestuurseenheid.uri}>.
 
-            {
-              SELECT ?fractie (COUNT(?_persoon) AS ?fractie_grootte)
-              WHERE {
-                ?_mandataris a mandaat:Mandataris.
-                ?_mandataris org:hasMembership/org:organisation ?fractie.
-                ?_mandataris org:holds ?_mandaat.
-                ?_mandaat org:role <${BESTUURSFUNCTIE_CODES.GEMEENTERAADSLID}>.
```
piemonkey, line 1679: NOTE! This was using `BESTUURSFUNCTIE_CODES.GEMEENTERAADSLID` for an RMW query. I corrected this to `LID_BCSD`. Please verify if this is correct.

abeforgit (commented):
trying this on a recent prod backup breaks the sync for me for mechelen
looks like a timeout thing as one of the queries is hanging

so if there's any way you can optimise, it would be worth trying

abeforgit:
ok well that last commit made the queries really fast :stuck_out_tongue:

abeforgit:
> > ok well that last commit made the queries really fast 😛
> 
> I still see a couple that are a little slow, but not as bad as before. Do we have a good way to test if this will be a problem on prod?

basically, no, but prod is quite beefy and this is a manually triggered thing with decent loading state handling so it should be ok

abeforgit (changes requested):
hmm it does seem like it's breaking the tables in the RVMW meeting of mechelen
on master: tables stay as is
on this branch: some tables end up empty

elpoelma (changes requested):
We do not have election results linked to OCMW 'bestuursorganen', so we can't do that ordering there. This is why these queries yield an empty result-set.

The gemeenteraad queries work as expected, fractions are now also ordered by amount of votes.

elpoelma:
> > We do not have election results linked to OCMW 'bestuursorganen', so we can't do that ordering there. This is why these queries yield an empty result-set.
> 
> Huh, was about to say the results seem to be missing (I also had to update my backup just to get results for master). Does this mean I should remove this ordering for now, or will we ever get results?

We will never get any I fear, I think it's best to simply remove the ordering now.

piemonkey:
I've removed the vote sorting from the RMW queries.

PR #787, opened 2024-12-04
Add configuration per administrative unit

lagartoverde:
### Overview
Add a way of adding configuration per administrative unit, in the future this will be replaced by user preferences and a call to the backend.
Also added the config needed to test https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/515

GN-5235
https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/515

### Setup
None

### How to test/reproduce
Login as Gemeente Gent, notice the style of the article titles is the old one. Then login as any other admin unit and notice we have the new article titles in there

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (changes requested):
Looks good overall!
- There's a high chance OCMW Gent want this change as well, so we probably also want to add a config for them
- I know we do it in quite a few places, but I'm more of a fan to use URI's to identify resources instead of their IDs. URIs should be the same on every environment, but IDs could (technically) differ
- In the future we will probably want to move these configuration settings to a seperate ember-service (as more options will come and users will be able to edit them). Not needed now though.

elpoelma (approved):

PR #789, opened 2024-12-04
Feat/wrap dowloaded docs for utf8

lagartoverde:
### Overview
Wrap the downloaded docs in correct html so we can specify in the head that is utf-8

Solved GN-5286

### Setup
None

### How to test/reproduce
Hard to reproduce but if you go to a prod copy of the Gemeente Oostkamp and try to download the agendapoint 1 of their IV the name should of the Tjörven mandatee should be correct

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/utils/wrap-downloaded-document.js
+++ b/app/utils/wrap-downloaded-document.js
@@ -1,0 +1,13 @@
+export function wrapDownloadedDocument(html) {
+  const document = `
+      <html>
+        <head>
+          <meta charset="utf-8" />
+        </head>
+        <body>
+          ${html}
+        </body>
+      </html>
+  `;
+  return document;
+}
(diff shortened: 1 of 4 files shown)


app/utils/wrap-downloaded-document.js
```diff
@@ -0,0 +1,13 @@
+export function wrapDownloadedDocument(html) {
+  const document = `
+      <html>
```
abeforgit, line 3: might as well be proper here and add the doctype? https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started#anatomy_of_an_html_document

abeforgit (changes requested):

abeforgit (approved):

PR #793, opened 2024-12-06
GN-5245: apply new template instantiate method on built-in standard templates

elpoelma:
### Overview
This PR ensures that we also apply the new `instantiateUuids` method when instantiating the built-in standard templates.

This PR should be backwards compatible and should thus also work with current versions of the backend.

[GN-5245](https://binnenland.atlassian.net/browse/GN-5245)
To be tested in conjunction with https://github.com/lblod/app-gelinkt-notuleren/pull/223

### Setup
Check-out https://github.com/lblod/app-gelinkt-notuleren/pull/223

### How to test/reproduce
See https://github.com/lblod/app-gelinkt-notuleren/pull/223 for instructions on how to test

### DIFF
--- a/.changeset/strong-rats-shave.md
+++ b/.changeset/strong-rats-shave.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Apply new `instantiateUuids` method onto built-in standard templates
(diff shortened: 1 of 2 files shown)


lagartoverde (approved):

piemonkey:
Doesn't https://github.com/lblod/app-gelinkt-notuleren/pull/223 make this redundant as we're replacing the standard templates with ones that don't use the `generateUuids` approach that is handled by `instantiateUuids`?

elpoelma:
> Doesn't [lblod/app-gelinkt-notuleren#223](https://github.com/lblod/app-gelinkt-notuleren/pull/223) make the `instantiateUuids` call redundant as we're replacing the standard templates with ones that don't use the `generateUuids` approach that is handled by `instantiateUuids`?

True, but I left it in case we e.g. accidentally deploy that frontend without running the migrations. I can also remove it of course.

PR #796, opened 2024-12-12
Add copy parts to the agendapoint

lagartoverde:
### Overview
Add a way of copying the parts of an agendapoint direclty from the agendapoint and not from the meeting

GN-5321

### Setup
None

### How to test/reproduce
Go to an agendapoint, in the file options you should have a copy decision parts button that should lead you to a new page where you can copy the decision parts

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/quiet-news-compete.md
+++ b/.changeset/quiet-news-compete.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': minor
+---
+
+Add copy parts link to the agendapoint file options
--- a/app/components/decision-copy.gjs
+++ b/app/components/decision-copy.gjs
@@ -1,0 +1,71 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { htmlSafe } from '@ember/template';
+import { generateExportTextFromEditorDocument } from 'frontend-gelinkt-notuleren/utils/generate-export-from-editor-document';
+import { copyStringToClipboard } from 'frontend-gelinkt-notuleren/utils/copy-string-to-clipboard';
+import AuToolbar from '@appuniversum/ember-appuniversum/components/au-toolbar';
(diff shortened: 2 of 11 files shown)


piemonkey (changes requested):
I feel like it doesn't make sense to duplicate this route (also template and controller, etc). To me it would make sense to have both sources use this new route instead of the old one. We'd then need to use a similar approach to when editing an agendapoint, where we use a query param to say which 'go back to' we display.

abeforgit:
> I feel like it doesn't make sense to duplicate this route (also template and controller, etc). To me it would make sense to have both sources use this new route instead of the old one. We'd then need to use a similar approach to when editing an agendapoint, where we use a query param to say which 'go back to' we display.

No I disagree. Duplicating the route is OK, and actually I think the query param shenanigans we do for the meeting are really bad. However, that doesn't mean we have to duplicate the _contents_ of the route
what this looks like it needs is to extract (most) of the copy page into a component, and then reference it in both routes

abeforgit:
ember routing fundamentally drives the logic of the app, it's (perhaps unfortunately) a bit more than just "this url will render this page"
I think the way they want you to think about routing is very "object oriented"
so two different ways of getting to the same page can have different routes imo, even if only to avoid the weird query param stuff

we have components to avoid duplication

abeforgit (changes requested):
imo: keep the route, but avoid the duplication by extracting to a component (but don't move the data-loading logic out of the route, just pass it through instead of loading it in the component)

elpoelma (approved):
Keeping two different routes and defining the logic in a single component indeed seems best. Looks good to me! I can easily access the copy-page using the two approaches.

PR #799, opened 2024-12-30
Feat/full rb copy

lagartoverde:
### Overview
Added copy screen to rb

GN-5300

### Setup
None

### How to test/reproduce
Go to regulatory attachments, notice you can now click on file options, copy parts, and the copy screen works as expected

### Challenges/uncertainties
In the config, the parts needs to be defined in order of specifity, didn't know how to solve this

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey (changes requested):
I'm not 100% on the approach. I think it depends how much this is intended to be a thing to keep around or whether it's a relatively quick fix. I say this as my original version was intended as a quick fix to support exporting IV parts to paste into vendor software.

I found a couple of issues:

- Somehow, navigating back from the RS copy route doesn't work, just the location changes but the page does not
- It seems like just nesting an article in a chapter doesn't work. I need to insert a section and subsection too.

abeforgit:
> I'm not 100% on the approach. I think it depends how much this is intended to be a thing to keep around or whether it's a relatively quick fix. I say this as my original version was intended as a quick fix to support exporting IV parts to paste into vendor software.

@piemonkey 

So we talked this through and considered other approaches, all involving adding some extra knowledge into the nodespecs so we can more robustly parse them.

While it would work, we ultimately hit a wall of flexibility. E.g.: we want _some_ block-rdfa nodes to show up in these views, but not all. While we could then make that configurable on the template level, I feel that's really going too far with this feature.

Personally I'd like to see the need for this page disappear eventually, maybe by moving the copy buttons really _into_ the editor or making the default copy smarter, or just approach the actual problem that these pages solve in a different way (more direct integration with vendors or something)

I was already annoyed at the demand for adding the copy page to RBs as it seemed like a whim of the PMs without much thought into who is actually supposed to use them.

In short, I think its best to go ahead with the slightly hacky html selectors, they're error prone but also very flexible, and we can drop them on a dime when we get the chance

abeforgit (changes requested):
still has the styling issue:

image

lagartoverde:
Sorry I thought I had paragraphs in my tests, seems like I didn't after the latest changes

translations/nl-BE.yaml
```diff
@@ -816,6 +817,7 @@ copy-options:
   heading: 'Kopieer onderdelen uit:'
   disclaimer: 'De opmaak van het document zal afhangen van de applicatie waarin het geüpload wordt.'
   copy-all: Kopieer volledig besluit
+  copy-rb: Copy full reglementaire bijlagen
```
elpoelma, line 820: This should be: 'Kopieer volledige reglementaire bijlage'

elpoelma (changes requested):
Functionally, this PR makes sense and works well.
There are still some translation-related issues left.

The 'Copy title' button is a bit weird to me, as you're actually copying the content below the title, but that's also due to how weirdly a 'title' is implemented/defined.

PR #807, opened 2025-01-13
Update to ember 5.12

piemonkey:
### Overview
Update packages to align with the blueprint, fix the linting, but don't fix the deprecations. Some deprecation fixes will follow in a later PR.

Editor PR: https://github.com/lblod/ember-rdfa-editor/pull/1237
Plugins PR: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/528
Jira ticket: https://binnenland.atlassian.net/browse/GN-5329

### Setup
N/A

### How to test/reproduce
Check that everything works!

### Challenges/uncertainties
Without updating ember-data, had to keep an older @ember/string as well as adding an @ember-data/store dep to prevent npm from trying to install v5.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [ ] no new deprecations


abeforgit (approved):
can't find any problems, looks good to me!

elpoelma (changes requested):
Found one issue (related to the usage of the removed `replaceRoute` method defined on controllers).
Relevant issue: https://github.com/emberjs/ember.js/issues/19497
(`this.router.replaceWith` fires the model hook unnecessarily)
We could introduce the suggested solution: https://github.com/emberjs/ember.js/issues/19497#issuecomment-2164641248

piemonkey:
> Found one issue (related to the usage of the removed `replaceRoute` method defined on controllers). Relevant issue: [emberjs/ember.js#19497](https://github.com/emberjs/ember.js/issues/19497) (`this.router.replaceWith` fires the model hook unnecessarily) We could introduce the suggested solution: [emberjs/ember.js#19497 (comment)](https://github.com/emberjs/ember.js/issues/19497#issuecomment-2164641248)

Hmm, I wasn't sure about replacing this. It was on my list of 'longer to fix ember 5 deprecations', one of the few that I didn't get to. What I was afraid of is that since we were already using `replaceRoute()`, we might be accidentally relying on this behaviour, so replacing it might break things. We definitely should fix it, with careful testing. Do you think it blocks this PR though?

elpoelma:
> > Found one issue (related to the usage of the removed `replaceRoute` method defined on controllers). Relevant issue: [emberjs/ember.js#19497](https://github.com/emberjs/ember.js/issues/19497) (`this.router.replaceWith` fires the model hook unnecessarily) We could introduce the suggested solution: [emberjs/ember.js#19497 (comment)](https://github.com/emberjs/ember.js/issues/19497#issuecomment-2164641248)
> 
> Hmm, I wasn't sure about replacing this. It was on my list of 'longer to fix ember 5 deprecations', one of the few that I didn't get to. What I was afraid of is that since we were already using `replaceRoute()`, we might be accidentally relying on this behaviour, so replacing it might break things. We definitely should fix it, with careful testing. Do you think it blocks this PR though?

I think the main issue is that the method has been removed in ember.js 5+, so it simply no longer works.

piemonkey:
> I think the main issue is that the method has been removed in ember.js 5+, so it simply no longer works.

Ah, then yes, we definitely should fix it...

elpoelma (approved):

PR #808, opened 2025-01-15
Fix deprecations from Ember 5 upgrade

piemonkey:
### Overview
I actually managed to fix a lot of deprecations, possibly all of them (except for the use of the `action` helper and modifier). It is possible that we could move to ember-data 5, but I haven't tried that yet.

Based on: https://github.com/lblod/frontend-gelinkt-notuleren/pull/807
Editor PR: https://github.com/lblod/ember-rdfa-editor/pull/1237
Plugins PR: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/528
Jira ticket: [binnenland.atlassian.net/browse/GN-5329](https://binnenland.atlassian.net/browse/GN-5329)

### Setup
It's possibly best to test with the `EXTEND_PROTOTYPES` flag set to false, just be aware that ember-drag-drop won't work, so that will need to be tested with it not set.

### How to test/reproduce
Test that everything works!

### Challenges/uncertainties
There's a lot of space for subtly different behaviour.

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/components/agenda-manager/agenda-item-form/select-location.js
+++ b/app/components/agenda-manager/agenda-item-form/select-location.js
@@ -16,9 +16,11 @@ export default class AgendaManagerAgendaItemFormSelectLocationComponent extends 
   }
 
   get afterItemOptions() {
-    return this.args.agendaItems
-      .sortBy('position')
-      .reject((x) => x === this.args.currentItem);
+    const agendaItems = this.args.agendaItems.filter(
+      (x) => x.id !== this.args.currentItem.id,
+    );
+    agendaItems.sort((a, b) => a.position - b.position);
+    return agendaItems;
   }
 
   @action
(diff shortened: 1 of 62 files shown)


app/components/agenda-manager/agenda-item-form/select-location.js
elpoelma, line 1: There seems to be an issue with this component. When I click open the dropdown to select an agendapoint, it immediately closes.

https://github.com/user-attachments/assets/8e5de8dd-e12f-4896-a990-1e31772bc015

elpoelma (changes requested):
Looks good overall, just found one issue.

app/components/agenda-manager/agenda-item-form/select-location.js
piemonkey: Huh. I tested that component pretty extensively as it had some issues. I'll try to reproduce. What browser were you on? Looks like FF, which is what I was mostly testing on...

elpoelma: I was indeed testing it with FF

elpoelma (approved):

PR #821, opened 2025-01-29
Use in element to fix rb copy parts

lagartoverde:
### Overview
Rb copy parts was working by accident on ember 4 and upgrading to ember 5 uncovered that some asumptions I made were plain wrong. This PR fixes all the faulty logic and also a missing translation I discovered on the demo

Didn't find it will link later

### Setup
None

### How to test/reproduce
Check that rb still works as expected in ember 5

### Challenges/uncertainties
none

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/cold-toes-happen.md
+++ b/.changeset/cold-toes-happen.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Fix rb copy parts in ember 5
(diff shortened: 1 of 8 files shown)


abeforgit (changes requested):
the button at the top says copy, but it downloads the document
image
otherwise looks good

lagartoverde:
True that was never working, I fixed that, sorry

PR #831, opened 2025-02-28
Add typescript support

abeforgit:
### Overview

Adds ts support, using the latest non-embroider blueprint as a reference. 
Also updates eslint to 9+ and flat mjs config
and as is tradition, moves one gjs component to gts to make sure it all works

### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (approved):
Looks good! :rocket:

piemonkey (commented):
Two small things I've noticed:
- TS doesn't accept `frontend-gelinkt-notuleren/` imports. I'm not sure if we care strongly about this though, we can just use relative imports.
- The types for `ember-data/store` don't seem to get picked up correctly.

elpoelma:
> Two small things I've noticed:
> 
>     * TS doesn't accept `frontend-gelinkt-notuleren/` imports. I'm not sure if we care strongly about this though, we can just use relative imports.
> 
>     * The types for `ember-data/store` don't seem to get picked up correctly.

I think we can solve the absolute `frontend-gelinkt-notuleren/` imports by adding them to the `paths` setting of the `tsconfig` file.

piemonkey:
@abeforgit I fixed the ember-data types and import path while working on the template picker, so I've pulled the TS changes into this PR.

piemonkey (approved):
There seems to be a problem with the eslint config as it seems to not pick up types for things we define in the `types` folder, when TS has no trouble with them. Not sure what's going on, but I don't think we need to wait for that to merge this.

PR #833, opened 2025-03-07
GN-5221: introduce application-wide `lang` query-param

elpoelma:
### Overview
This PR introduces an application-wide `lang` query parameter which controls the display language of the application. By default, the language of the application is set to `nl-BE`, regardless of the user navigator languages/system locale.

[GN-5221](https://binnenland.atlassian.net/browse/GN-5221?atlOrigin=eyJpIjoiZjgyZWMwNWE4Y2ZkNGY1Zjk1MDkzMzlmZGMwM2VkNzYiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the app
- Ensure the display language is dutch
- Add a `lang` query-param with e.g. `en-US`/`en-UK`
- Ensure the display language is now english
- The language should be persisted across reloads and route transitions
- Setting an unsupported language results in a dutch display language

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/empty-pens-drum.md
+++ b/.changeset/empty-pens-drum.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': minor
+---
+
+Addition of an application-wide `lang` query parameter which controls the display language of the application. By default, the language of the application is set to `nl-BE`, regardless of the user navigator languages/system locale.
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -5,7 +5,11 @@ import { decentLocaleMatch } from '../utils/intl';
 
 const featureFlagRegex = /^feature\[(.+)\]$/;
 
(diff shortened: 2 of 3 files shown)


lagartoverde (changes requested):
I would do this only in prod, as it will be annoying for devs

elpoelma:
> I would do this only in prod, as it will be annoying for devs

We could maybe introduce a `DEFAULT_LANGUAGE` env var which sets/overrides the default language.
We could then set this var to 'en-US' on DEV, and to 'nl-BE' on QA and PROD.

abeforgit:
> I would do this only in prod, as it will be annoying for devs

is it that bad? the query param persists across anything you do basically
you can even store it as a bookmark

that being said what @elpoelma proposed also seems sensible, if it's easy to add

abeforgit (approved):

PR #836, opened 2025-03-13
Move persisting documents into the document service

piemonkey:
### Overview
Git has done a decent job of spotting renames in this one, so easier to review, even with the typescript fun... The progressive typescript migration doesn't seem too tricky to do by pieces. 

Implements the template picker flow for document creation, it's missing the favourites functionality and the move of standard templates to live in RB, so the number of results on the first page of AP templates is wrong.

Draft until the plugins PR is merged and is based off the `internal/typescript` branch, so ideally that would be merged too.

Part (most) of [binnenland.atlassian.net/browse/GN-5418](https://binnenland.atlassian.net/browse/GN-5418)
Design: https://www.figma.com/design/bxuSHLkiCQEG9UVmcTIITi/AP_template?node-id=0-1&p=f
Uses: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/553

### Setup
N/A

### How to test/reproduce
Test creating new agendapoints or regulatory statements. Include picking decision types that include subtypes and sub-subtypes.

### Challenges/uncertainties
It feels like there was a lot of stuff hidden in this ticket...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (commented):
This looks really good!
Selecting a template and then the decision type + title is very intuitive and a nice improvement :)
Also like we're using typescript now :)

Some suggestions:
- Maybe we could prefill the decision type dropdown in case the template already has a decision type (e.g. when using the 'Aanvullend reglement besluit' template)
- Maybe add some margin on top of the 'Terug naar sjabloon kiezen' button

piemonkey:
@elpoelma I didn't realise some templates had built in types. I added detection logic, tweaked the button layout and updated to a released version of plugins.

lagartoverde:
image
Don't know if this can be fixed, as I remember having problems with the powerselect as well, but I think the dropdown should be on top of the modal and not need scrolling

piemonkey:
@lagartoverde turns out there's no need for fighting z-indexes, there's an overflow arg on AuModal. It's fixed now.

lagartoverde (approved):
Good job

PR #840, opened 2025-03-24
GN-5465: support for meeting validation

elpoelma:
## Overview
This PR implements a meeting validation feature. If there are any issues with the currently opened meeting, the user should see warning/error message prompting them to fix these.
What is included in the meeting validation:
- Validation of the meeting general information
- Validation of the meeting attendance
- Validation of the attendance of the agendapoint treatments

A card on the right side of the meeting form shows an overview of all problems with the meeting. I had to adjust the layout of the meeting a bit to accommodate for the new right sidebar. (It should be mostly responsive)

If e.g. the attendance section is not fully correct (e.g. non assigned mandatees), an alert is shown containing the details of the non-assigned mandatees.

[GN-5465](https://binnenland.atlassian.net/browse/GN-5465)

### Functionality
- Start the application and log in as a municipality
- Open a meeting
- On the right side of the meeting form, a card should be shown containing details of the meeting validity. Ensure these details seem/are correct
- Using the validity card, you should be able to navigate to the problematic sections
- When modifying/updating a section, the validity card should update.
- If a meeting attendance section is filled in, but there are non-assigned mandatees, a warning about this should be shown in the relevant section. To simulate this, you can remove a mandatee from the attendees/absentees from a meeting using SPARQL
- If an error occurs while computing the meeting validation, an error message is shown on the right side of the meeting
- Collapsed state:
  * The meeting has no problems? Validity card is collapsed by default
  * The meeting has problems? Validity card is expanded by default

### Layout/Responsiveness
- When opening the application on a laptop/desktop, the meeting form should be shown in combination with both the left and right sidebar
- When reducing the screen width, the left sidebar (as it is the least important one) is the first to disappear
- When reducing the screen width even more, the right sidebar goes into `position: absolute` mode and is overlaid on top of the right section of the meeting form

### Checks PR readiness
- [x] changelog
- [x] npm lint


abeforgit (changes requested):
The validation list also takes over the role of navigating the meeting document, since the left sidebar is now gone.

However, when you fully "complete" a decision by adding participants, it dissapears from the completion list (rather than getting a green checkmark), which takes away your ability to navigate and feels like you broke something.

elpoelma:
> The validation list also takes over the role of navigating the meeting document, since the left sidebar is now gone.
> 
> However, when you fully "complete" a decision by adding participants, it dissapears from the completion list (rather than getting a green checkmark), which takes away your ability to navigate and feels like you broke something.
> 
> image

The left sidebar should still be there, it is only hidden now on smaller screen sizes. I didn't fully intend the right sidebar as a 'navigation' bar, but rather as a 'problems'/'validation' bar (which users can also use to navigate to the sections with issues. I can of course add the other treatments (with no problems too), and just add a green checkmark next to them.

elpoelma:
@abeforgit I've now added all treatments to the validation card (error or not). I'm thinking we then might as well also add the other navigation items and just remove the navigation left sidebar altogether...

elpoelma:
@abeforgit what do you think of these changes? I still need to add a little message when hovering over the warning icons though. Css isn't great, so might need some improvements.

abeforgit (changes requested):
the validation/navigation element should be open by default. If we make it collapsible, it should save the collapsed state (sorry, this is just always the case if we make these kinds of UIs)

abeforgit (changes requested):
Also: we shouldn't block going to the publication screen when there are validation issues. It's necessary that they can publish the agenda before any of the meeting details have been filled out

elpoelma:
> Also: we shouldn't block going to the publication screen when there are validation issues. It's necessary that they can publish the agenda before any of the meeting details have been filled out

Fixed!

elpoelma:
> the validation/navigation element should be open by default. If we make it collapsible, it should save the collapsed state (sorry, this is just always the case if we make these kinds of UIs)

That makes sense, the navigation element being initially collapsed is still an artifact of when it was implemented only as a validation card, I'll adjust the behaviour.

abeforgit (changes requested):
Don't know if it's a regression in this PR, but there's a weird bug:

- make a new meeting
- add a new agenda item 
- add another agenda item (not a premade one, a new one from the meeting ui)

Something breaks and you get NaN as the title:
image

Weirdly enough, adding a third item (and subsequent ones) seems to work without issue, though I've also seen it not work for multiple ones

**Refreshing the page** fixes the problem. So it seems like a timing thing possibly

There's definitely some data issue and not just a rendering thing, as editing the item doesn't work either.

Console error:
image

abeforgit:
I also made a follow up story, as it now feels natural to add a few more validations:
https://binnenland.atlassian.net/browse/GN-5563

elpoelma:
> Don't know if it's a regression in this PR, but there's a weird bug:
> 
>     * make a new meeting
> 
>     * add a new agenda item
> 
>     * add another agenda item (not a premade one, a new one from the meeting ui)
> 
> 
> Something breaks and you get NaN as the title: image
> 
> Weirdly enough, adding a third item (and subsequent ones) seems to work without issue, though I've also seen it not work for multiple ones
> 
> **Refreshing the page** fixes the problem. So it seems like a timing thing possibly
> 
> There's definitely some data issue and not just a rendering thing, as editing the item doesn't work either.
> 
> Console error: image

should be fixed now!

abeforgit (approved):

PR #857, opened 2025-05-08
Feat/validate decision presence

lagartoverde:
### Overview
Validate decision presence in treatment. In this we convert the treatment to triples, which is a good thing since we will want to apply a shacl shape to that dataset at some point

GN-5563

### Setup
None

### How to test/reproduce
Check that if you add an agendapoint without a decision it show you a validation warning in the UI

### Challenges/uncertainties
Check the dutch translations

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/orange-months-beg.md
+++ b/.changeset/orange-months-beg.md
@@ -1,0 +1,5 @@
+---
+'frontend-gelinkt-notuleren': patch
+---
+
+Validate decision presence in treatment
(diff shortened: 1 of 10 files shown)


piemonkey (commented):
Looks good! One small thing though is that the tooltip for treatment problems needs to be updated to reflect the fact that we're not only looking at attendance:
image
It would also be nice to at least say "one or more", but better to have it reflect the actual number...

While you're tweaking translations, I spotted that one is missing too:
image

piemonkey (approved):

PR #910, opened 2026-04-21
Query categories in the AR importer so onderborden signs can be marked

lagartoverde:
### Overview
Categories were not pulled on the AR ontwerp plugin so it couldn't be marked correctly in the template

https://binnenland.atlassian.net/browse/GN-6076

### Setup
None

### How to test/reproduce
Go to an agendapoint, click on AR ontwerp and notice that if you include a onderborden sign it's marked correctly in the result template

### Challenges/uncertainties
Not sure about the flow of the AR, but I'm querying directly to MOW for the categories, could it be that we need to include the categories in the script that fetches this data?

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey (commented):
I think since we're querying MOW in the VKS design service, we should be [adding to the query there](https://github.com/lblod/vks-design-service/blob/master/lib/db/traffic-signal-concepts.ts). Doing it in the front-end means adding an additional query.

lagartoverde:
> I think since we're querying MOW in the VKS design service, we should be [adding to the query there](https://github.com/lblod/vks-design-service/blob/master/lib/db/traffic-signal-concepts.ts). Doing it in the front-end means adding an additional query.

Done :)

piemonkey (approved):
Seems to work well :)

PR #912, opened 2026-04-22
Add indication when an AR design only includes existing signals

piemonkey:
### Overview
Adds a ~~pill to the AR design~~ template comment to the preview of the AR design indicating when a ~~design~~ measure within a design only includes signals that are already existing. The template comments should not appear in the actual document.

Both which statuses to include in this check and how to display it (including translation) were not specified, so need to be checked.

Part of https://binnenland.atlassian.net/browse/GN-5954
Needs https://github.com/lblod/vks-design-service/pull/15

### Setup
Run against a GN including the VKS PR. To find some AR designs which only include existing signs, you can use the gemeente Hechtel-Eksel.

### How to test/reproduce
In an AR, open the modal to insert an AR design. Click on the preview for a design, if the signals included are all existing, then you should see a ~~pill next to the insert button~~ template comment before each measure indicating that.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
Looks good overall, but you went one level further than I think is expected (though yeah, none of this has been particularly thought through, you couldn't know)

I think based on "the vibes I was getting" from random meetings where this was never actually the topic, I think we'd want an indicator per-measure.

The idea is that AR documents are usually viewed on the level of a full street. 

So a municipality wants to organize their designs in VKS this way as well. 
This means that new users coming in to VKS that want to use it to add a new sign in a street, would like to make an AR design that contains all the current measures (aka signcombos) in that street, and then add the new one.

That's what the status indicates

piemonkey:
@abeforgit I've switched to per-measure indications now

translations/nl-BE.yaml
```diff
@@ -960,6 +960,7 @@ ar-importer:
   preview:
     return-to-overview: Terug naar overzicht
     insert: Voeg in
+    only-existing: Dit ontwerp omvat alleen bestaande Verkeersteken
```
abeforgit:
```suggestion
    only-existing: Dit maatregelontwerp bevat enkel reeds bestaande verkeerstekens.
```

app/services/ar-importer.ts
```diff
@@ -156,28 +158,54 @@ export default class ArImporterService extends Service {
...
+                },
+              ]
+            : []),
+          insertMeasure({
+            arDesignUri: design.uri,
+            measureDesign: {
+              uri: measureDesign.uri,
+              measureConcept: {
+                uri: measureConcept.uri,
+                label: `${measureConcept.label}${onlyExistingSignals ? this.intl.t('ar-importer.preview.only-existing') : ''}`,
```
abeforgit: needs a space, but also a simplified text, it's too long now

I think something like `(bestaand)` is enough

abeforgit (changes requested):

translations/nl-BE.yaml
piemonkey: ~~This is only used in the inline warning, so I guess it should just be `bestaand`?~~ Ignore this... I'll add a separate translation for just `bestaand`.

PR #915, opened 2026-04-23
Move document-service methods with side-effects to not use tasks

piemonkey:
### Overview
The issue here is that if the document creation process is interrupted, which is easy to do when the document is large, such as the regulatory statement with 60 articles. 
This aims to avoid issues with cancellation when navigating away and on reload but not errors.

The changes are made to both RSs and APs to keep behaviour consistent.

Jira ticket: [binnenland.atlassian.net/browse/GN-6079](https://binnenland.atlassian.net/browse/GN-6079)
Fix for crash: https://github.com/lblod/frontend-gelinkt-notuleren/pull/914

### Setup
See linked PR fixing the crash for data set-up.

### How to test/reproduce
All kinds of saving operations should still work (including editing titles and confirming simultaneous edits), as well as copying agendapoints. Loading messages should be shown at the right times. Navigating away, reloading or closing the tab, should result in confirmation dialogues if done while saving is in progress.

### Challenges/uncertainties
Unload confirmations mess with Ember dev reloads. It would be good if there was a way to detect these and not ask for confirmation.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (approved):
CFR detecting development mode, there's maybe something we can do with NODE_ENV=development?

additionally, there are checks like that in the environment.js file. We could add a value to the ember environment there and check for that perhaps?

piemonkey:
> CFR detecting development mode, there's maybe something we can do with NODE_ENV=development?
> 
> additionally, there are checks like that in the environment.js file. We could add a value to the ember environment there and check for that perhaps?

It would be easy to just detect `NODE_ENV=development`, but I'm always wary of doing things like that as it means you can miss problems with the logic that is being ignored as you're always testing in dev mode. It would be nice if ember either added something to the unload event or sent a custom event just before reloading...

PR #925, opened 2026-05-29
GN-6139: Implement insert options for AR-design insert

abeforgit:
### Overview

- adds the requested insertion options, using the same UI logic as was used in the roadsign plugin

**UX Change**

I also made it so a user always has to go to the preview first before inserting. It simplifies the UX, while only adding a click in cases where there are no warnings to show.

If I didn't do this, I'd have had to add the insertion selector both in the preview window and in the table, which would be very messy.

[GN-6139](https://binnenland.atlassian.net/browse/GN-6139)

### Setup

Doesn't need any special backend, so you can proxy to dev

### How to test/reproduce

Insert an AR, notice the "preview" and "insert" actions have been replaced by a single "select" action, which opens the preview. 
There, you can now see the same insertion options as we have in the roadsign plugin. 

The option list will populate with the amount of articles in the document.

### Checks PR readiness
- [x] UI: works on smaller screen sizes (until about 1/5 of the screen)
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -945,8 +945,7 @@ ar-importer:
         date: Last modified
         status: Usage status
       actions:
-        preview: Preview
-        insert: Insert
+        select: Select
       statuses:
         used: Used
         unused: Not yet used
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -955,6 +955,7 @@ ar-importer:
     first: Before the first article
     last: After the last article
     after-article-x: After article {articleNumber}
+    select-position: Choose insertion point
   preview:
     return-to-overview: Return to overview
     only-existing-label: existing
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -951,9 +951,13 @@ ar-importer:
         used: Used
         unused: Not yet used
       no-data: No designs found
+  controls:
+    insert: Insert
+    first: Before the first article
+    last: After the last article
+    after-article-x: After article {articleNumber}
   preview:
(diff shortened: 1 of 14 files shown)


kobemertens:
image
Getting a visual error where the button goes out of view. Code looks good though! :ok_hand: Nice refactor

abeforgit:
> image Getting a visual error where the button goes out of view. Code looks good though! 👌 Nice refactor

wut, I just tested this, will recheck

translations/en-us.yaml
```diff
@@ -945,15 +945,19 @@ ar-importer:
...
       statuses:
         used: Used
         unused: Not yet used
       no-data: No designs found
+  controls:
+    insert: Insert
+    first: Before the first article
+    last: After the last article
+    after-article-x: After article {articleNumber}
+    select-position: Choose insertion point
```
kobemertens, line 958: Why not reuse the keys I added for the mobility measure insertion options? Also they are different from those.

kobemertens:
image

Maybe not really part of this PR, so feel free to ignore. But I would put the insert button on the bottom right, which follows the natural reading direction more.

translations/en-us.yaml
abeforgit: I don't think you can import translation keys? and if you could, not sure if I'd want to have dependencies entangled like that

abeforgit:
> image
> Maybe not really part of this PR, so feel free to ignore. But I would put the insert button on the bottom right, which follows the natural reading direction more.

done!

PR #931, opened 2026-06-30
Allow template selection from meeting edit

kobemertens:
### Overview
Until now it was not possible to select a template when using the edit agendapoint content button on the meeting edit page. This PR shows the template selector when the agendapoint document has no content yet.

https://binnenland.atlassian.net/browse/GN-6166

### How to test/reproduce
Add an agendapoint via the meeting edit page. Click the "bewerk" button on its content, the template selector modal should appear.

### Challenges/uncertainties
The modal is now opened when a agendapoint document has not content yet, this feels like a heuristic but it should work most of the time.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/controllers/meetings/edit/agendapoint/new.ts
+++ b/app/controllers/meetings/edit/agendapoint/new.ts
@@ -33,7 +33,7 @@ export default class MeetingsEditNewController extends Controller {
     void this.plausible.trackEvent('Create agendapoint', {
       templateTitle: chosenTemplate.title,
     });
-    this.router.transitionTo('agendapoints.edit', container.id);
+    this.router.transitionTo('meetings.edit.agendapoint.edit', container.id);
   }
 
   get editorConfig() {
--- a/app/controllers/meetings/edit/agendapoint/new.ts
+++ b/app/controllers/meetings/edit/agendapoint/new.ts
@@ -1,0 +1,78 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { service } from '@ember/service';
+import type RouterService from '@ember/routing/router-service';
+import type IntlService from 'ember-intl/services/intl';
+import type { SayController } from '@lblod/ember-rdfa-editor';
+import type MeetingsEditRoute from 'frontend-gelinkt-notuleren/routes/meetings/edit';
+import type { ModelFrom } from 'frontend-gelinkt-notuleren/utils/types';
+import type TemplateFetcher from 'frontend-gelinkt-notuleren/services/template-fetcher';
+import { EDITOR_FOLDERS } from 'frontend-gelinkt-notuleren/config/constants';
+import type AgendapointEditorService from 'frontend-gelinkt-notuleren/services/editor/agendapoint';
+import type DocumentContainerModel from 'frontend-gelinkt-notuleren/models/document-container';
+import type { Template } from 'frontend-gelinkt-notuleren/services/template-fetcher';
+import type Plausible from 'ember-plausible/services/plausible';
+
+export default class MeetingsEditIntroController extends Controller {
+  declare model: ModelFrom<MeetingsEditRoute>;
+  @service declare router: RouterService;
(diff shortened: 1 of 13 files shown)


app/controllers/meetings/edit/agendapoint/new.ts
```diff
@@ -0,0 +1,80 @@
...
+  @action
+  redirectToAgendapoint(
+    container: DocumentContainerModel,
+    chosenTemplate: Template,
+  ) {
+    // Plausible Analytics: post custom event about the template used to create the agendapoint
+    void this.plausible.trackEvent('Create agendapoint', {
+      templateTitle: chosenTemplate.title,
+    });
+    this.router.transitionTo('agendapoints.edit', container.id);
```
piemonkey: This shouldn't be the `agendapoints.edit` route, as this pulls the user out of the meeting creation flow. It should (I think) be the `meetings.edit.agendapoint.edit` route.

app/controllers/meetings/edit/agendapoint/new.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  get editorConfig() {
+    return this.agendapointEditor.config;
+  }
+
+  getTemplates = async ({
```
piemonkey, line 59: It would be nice to pull this out to a separate function as it's used in 3 places.

piemonkey (commented):
Looks good, just the route used for editing the AP needs to be fixed.

There are some issues that came up in testing, but I think these already exist.

1. It is possible to avoid the template picker by creating an AP as a meeting agenda item, then going back to the AP list and editing the AP.
2. If you create an AP as a meeting agenda item, then return to the AP list without 'creating' it, it is listed as `geagendered` (as with 'meet create 2' in the image). If you create an AP as a meeting agenda item, then create it using the new flow, when you go to the AP list it is listed as `in voorbereiding`.
Image
3. If you create an AP as a meeting agenda item, then in the new 'create' flow, change the title of the AP, this is not reflected in the agenda of the meeting. It is possible to update the description of the meeting from the doc in the 'edit agenda item' modal, but not the title.

4. It seems that when creating an AP in the meeting edit page, then use the create flow to make the document, you can later add the AP to the agenda a second time. After the AP has been created but the document not created, it is (correctly) not possible to do this.

app/controllers/meetings/edit/agendapoint/new.ts
abeforgit: indeed, solved now

abeforgit:
> 1. It is possible to avoid the template picker by creating an AP as a meeting agenda item, then going back to the AP list and editing the AP.

I think I can live with that for now

> 2. If you create an AP as a meeting agenda item, then return to the AP list without 'creating' it, it is listed as `geagendered` (as with 'meet create 2' in the image). If you create an AP as a meeting agenda item, then create it using the new flow, when you go to the AP list it is listed as `in voorbereiding`.

this is fixed as of [e5d5a3d](https://github.com/lblod/frontend-gelinkt-notuleren/pull/931/commits/e5d5a3d6675b3373a38511989b62b8b4cf135139)


> 4. It seems that when creating an AP in the meeting edit page, then use the create flow to make the document, you can later add the AP to the agenda a second time. After the AP has been created but the document not created, it is (correctly) not possible to do this.

this was the same bug as `2`, so is now also fixed

looking at nr 3 now

abeforgit:
> 3. If you create an AP as a meeting agenda item, then in the new 'create' flow, change the title of the AP, this is not reflected in the agenda of the meeting. It is possible to update the description of the meeting from the doc in the 'edit agenda item' modal, but not the title.

this is now fixed as of [56b356e](https://github.com/lblod/frontend-gelinkt-notuleren/pull/931/commits/56b356e939c6726d12c61179ced2ac3cadb2b5e3)

abeforgit:
@kobemertens fixed a few things, would be good if you could have a look in the morning
(the fix in 57d8fc1 is not because of this pr, but was always the case with documents coming from the documentcreator)

PR #932, opened 2026-07-24
Feature/agencies

kobemertens:
### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


kobemertens:
Apparently this was because the `skos` prefix was missing in one of the migrations which caused it to fail... Not sure why it worked on my end though.

